PR #12, opened 2019-05-17
Multiple files support

moshababo:
Adding multiple files support, based on https://github.com/spacemeshos/merkle-tree/pull/8.

This is needed for multiple features: sub-ids (multiple units), concurrent execution, etc.


go.sum
```diff
@@ -94,6 +95,16 @@ github.com/spacemeshos/go-spacemesh v0.0.0-20190218131258-f9290a04382c h1:F2uZn5
...
+github.com/spacemeshos/merkle-tree v0.0.0-20190515130439-516ccde24669/go.mod h1:mPxjt4RONPxSUhxOq4bhSJyKVGQJ0VMSyRiE51dDLgE=
+github.com/spacemeshos/merkle-tree v0.0.0-20190516131630-fa084e188123 h1:KwJBZc1biFWG2M+G/1brjEch+3x8c1oFaEmXRZRAsxc=
+github.com/spacemeshos/merkle-tree v0.0.0-20190516131630-fa084e188123/go.mod h1:mPxjt4RONPxSUhxOq4bhSJyKVGQJ0VMSyRiE51dDLgE=
+github.com/spacemeshos/merkle-tree v0.0.0-20190516142306-14d56eab8777 h1:PziGbeGPTsimJCmZ93Ae8/wEenSo6X8jptgTGJHV8I8=
+github.com/spacemeshos/merkle-tree v0.0.0-20190516142306-14d56eab8777/go.mod h1:mPxjt4RONPxSUhxOq4bhSJyKVGQJ0VMSyRiE51dDLgE=
+github.com/spacemeshos/merkle-tree v0.0.0-20190516182455-58d29a79ffa5 h1:fHQs3bA3rTcZ+HZFJQn4oU4DwftxMRFWnu+He0LNCDo=
+github.com/spacemeshos/merkle-tree v0.0.0-20190516182455-58d29a79ffa5/go.mod h1:mPxjt4RONPxSUhxOq4bhSJyKVGQJ0VMSyRiE51dDLgE=
+github.com/spacemeshos/merkle-tree v0.0.0-20190517113914-1670e9ad4d58 h1:uHKiHRdazgCkYYNkzRwLAbmSuoogNao6Mz/u8H7gPZw=
+github.com/spacemeshos/merkle-tree v0.0.0-20190517113914-1670e9ad4d58/go.mod h1:mPxjt4RONPxSUhxOq4bhSJyKVGQJ0VMSyRiE51dDLgE=
```
noamnelke, line 97: looks like a `go mod tidy` is in order...

proving/space.go
```diff
@@ -1,27 +1,44 @@
...
 
-// LabelGroups returns the number of label groups of a given space amount.
-func (s Space) LabelGroups(labelGroupSize uint64) uint64 {
-	return uint64(s) / labelGroupSize
+func NumOfFiles(space uint64, filesize uint64) (int, error) {
+	if err := ValidateFileSize(space, filesize); err != nil {
+		return 0, err
+	}
+
+	return int(space / filesize), nil
```
noamnelke, line 38: Rounding down means the last file will be a tree with the same height as the rest, but a greater size. This will technically work, but as a user, I'd expect the defined filesize to be a maximum size, not a minimum, so this behaviour would surprise me.

I would round up, but up to you.

noamnelke (approved):
Great changes 👏🏻

proving/space.go
moshababo: Right, but uneven files distribution is not supported yet, and we validate it [here](https://github.com/spacemeshos/post/pull/12/files#diff-a1409f30b2a179b5e8d9dc21ef00376fR20).
So i'll leave it as-is until it will be supported.

PR #13, opened 2019-05-20
Parallelism support

moshababo:
Adding parallelism support, based on https://github.com/spacemeshos/post/pull/12.

Number of workers (goroutines) is: `min(num of cpus, num of files)`.
Number of files is not adjusted yet. Used only in tests.

Basic _initialization_ benchmarks:
````
1.1GB space
32 files
12 CPUs (6 physical)

no parallelism: 235 sec
no parallelism 1 file: 243 sec
parallel: workers as num of files (32): 64 sec
parallel: workers as num of cpus (12): 63 sec
````
````
2.2GB space
32 files
12 CPUs (6 physical)

no parallelism:  476 sec
no parallelism 1 file: 501 sec
parallel: workers as num of files (32): 128 sec
parallel: workers as num of cpus (12): 131 sec
parallel: workers as num of cpus / 2 (6): 137 sec
parallel: workers as num of cpus / 4 (3): 210 sec
````

### DIFF
--- a/initialization/initialize.go
+++ b/initialization/initialize.go
@@ -9,6 +9,7 @@ import (
 	"github.com/spacemeshos/post/persistence"
 	"github.com/spacemeshos/post/proving"
 	"github.com/spacemeshos/post/shared"
+	"runtime"
 )
 
 const (
@@ -26,11 +27,11 @@ type (
 // Difficulty determines the number of bits per label that are stored. Each leaf in the tree is 32 bytes = 256 bits.
 // The number of bits per label is 256 / LabelsPerGroup. LabelsPerGroup = 1 << difficulty.
 // Supported values range from 5 (8 bits per label) to 8 (1 bit per label).
-func Initialize(id []byte, space uint64, numOfProvenLabels uint8, difficulty proving.Difficulty) (*proving.Proof, error) {
-	return initialize(id, space, space, numOfProvenLabels, difficulty)
+func Initialize(id []byte, space uint64, numOfProvenLabels uint8, difficulty proving.Difficulty, parallel bool) (*proving.Proof, error) {
+	return initialize(id, space, space, numOfProvenLabels, difficulty, parallel)
 }
 
-func initialize(id []byte, space uint64, filesize uint64, numOfProvenLabels uint8, difficulty proving.Difficulty) (*proving.Proof, error) {
+func initialize(id []byte, space uint64, filesize uint64, numOfProvenLabels uint8, difficulty proving.Difficulty, parallel bool) (*proving.Proof, error) {
 	if err := proving.ValidateSpace(space); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by moshababo** [comment]
`GOMAXPROCS` will be set after implementing https://github.com/spacemeshos/go-spacemesh/issues/907.

File: initialization/initialize.go [hunk 1]
```diff
@@ -89,9 +85,63 @@ type initResult struct {
...
-func initializeChunk(id []byte, chunkPosition int, labelGroupsPerChunk uint64, difficulty proving.Difficulty) (*initResult, error) {
+type initChunkResult struct {
+	index int
+	*initResult
+}
+
+func initializeChunks(id []byte, difficulty proving.Difficulty, numOfChunks int, labelGroupsPerChunk uint64, parallel bool) ([]*initResult, error) {
+	var numOfWorkers int
+	if parallel {
+		numOfWorkers = min(runtime.NumCPU(), numOfChunks)
```
zalmen, line 96: `runtime.NumCPU` gives you the number of logical cores, I think the relevant number should be the number of physical cores as the work is cpu intensive. What do you think?
Also, not sure why, but when I run `runtime.NumCPU` on my mac I get 1, what am I missing?

initialization/initialize.go
```diff
@@ -26,11 +27,11 @@ type (
 // Difficulty determines the number of bits per label that are stored. Each leaf in the tree is 32 bytes = 256 bits.
 // The number of bits per label is 256 / LabelsPerGroup. LabelsPerGroup = 1 << difficulty.
 // Supported values range from 5 (8 bits per label) to 8 (1 bit per label).
-func Initialize(id []byte, space uint64, numOfProvenLabels uint8, difficulty proving.Difficulty) (*proving.Proof, error) {
-	return initialize(id, space, space, numOfProvenLabels, difficulty)
+func Initialize(id []byte, space uint64, numOfProvenLabels uint8, difficulty proving.Difficulty, parallel bool) (*proving.Proof, error) {
```
zalmen, line 30: I know it is not part of this PR, but I think that instead of passing the space size we should provide the number of units where the size of the unit is configurable

initialization/initialize.go
```diff
@@ -26,11 +27,11 @@ type (
...
+	return initialize(id, space, space, numOfProvenLabels, difficulty, parallel)
```
zalmen, line 31: I don't understand the `filesize` param and why it is set to `space`

initialization/initialize.go
```diff
@@ -89,9 +85,63 @@ type initResult struct {
...
+		}()
+	}
+
+	results := make([]*initResult, numOfChunks)
+	for i := 0; i < numOfChunks; i++ {
+		select {
+		case res := <-resultsChan:
+			results[res.index] = res.initResult
+		case err := <-errChan:
+			return nil, err
```
zalmen, line 135: We need to have a proper teardown in case of failure, we shouldn't leave stale init files

zalmen (changes requested):

initialization/initialize.go
moshababo: Although it make sense, I’d rather do better benchmarking first.

initialization/initialize.go
moshababo: I wanted to do this also to `numOfProvenLabels`, by keeping these params only in a private function which is needed for testing (same as `filesize`). It wasn’t straight forward due to some dependencies. 

In the end we’ll expose only what we want the client to dynamically control.

initialization/initialize.go
moshababo: As mentioned above, having `filesize` as a param is needed for testing. 
I thought not to expose `filesize` to the client, so it’s not a param for the public function.
In addition, I didn’t want to deal with the `filesize` calculation logic right now, so i’m merely using `space`, and there will be only 1 file (I can also use `numOfFiles` instead).

initialization/initialize.go
moshababo: I'll rather deal with the teardown in my other PR - don't want to create additional merge problems.

initialization/initialize.go
moshababo: If we'll want this package to be consumed directly by the client (the fullnode) and also by `rpc` package, it's a bit more problematic, and the best would probably be to have an initialization instance which contains config.

initialization/initialize.go
zalmen: NP, it is best to have the level of concurrency from the configuration

zalmen (approved):

PR #14, opened 2019-05-23
Post server

moshababo:
Closing https://github.com/spacemeshos/go-spacemesh/issues/907.

Fixed:
* server/package mode support
* viper/smutil-log usage
* async init/exec api calls support


config.go
```diff
@@ -0,0 +1,244 @@
...
+	defaultLogDirname     = "logs"
+	defaultLogFilename    = "post.log"
+	defaultMaxLogFiles    = 3
+	defaultMaxLogFileSize = 10
+	defaultRPCPort        = 50001
+	defaultRESTPort       = 8080
+)
+
+var (
+	defaultPostDir    = btcutil.AppDataDir("post", false)
```
zalmen: It is inconsistent with how we set the state directory in other modules. It seems that this method might be a good approach, but if we choose to use it we should be consistent and use only it.

config.go
```diff
@@ -0,0 +1,244 @@
+package main
```
zalmen: Why not use the viper config infra?

initialization/initialize.go
```diff
@@ -169,3 +173,16 @@ func merge(results []*initResult) (*initResult, error) {
 		return &initResult{reader, root}, nil
 	}
 }
+
+func Reset(dir string) (*persistence.ResetResult, error) {
```
zalmen: Add documentation (as with all exported methods)

initialization/initialize.go
```diff
@@ -169,3 +173,16 @@ func merge(results []*initResult) (*initResult, error) {
...
+	res, err := persistence.Reset(dir)
```
zalmen: This call should also stop init execution, in case still running? If yes, add a todo and we'll handle it once we add the stop/resume functionallity.

rpc/api/api.proto
```diff
@@ -0,0 +1,114 @@
+syntax = "proto3";
+
+import "google/api/annotations.proto";
+
+package api;
+
+service Post {
```
zalmen, line 7: What about a validate call? (for testing purposes)

rpc/rpcserver.go
```diff
@@ -0,0 +1,170 @@
...
+func NewRPCServer(s *signal.Signal, params *shared.Params, datadir string, lograte uint64) *rpcServer {
+	return &rpcServer{
+		s:       s,
+		params:  params,
+		datadir: datadir,
+		lograte: lograte,
+	}
+}
+
+func (r *rpcServer) Initialize(ctx context.Context, in *api.InitializeRequest) (*api.InitializeResponse, error) {
```
zalmen, line 55: Initialize should be non-blocking. It should return once it validated the input.

rpc/rpcserver.go
```diff
@@ -0,0 +1,170 @@
...
+		Challenge:    proof.Challenge,
+		MerkleRoot:   proof.MerkleRoot,
+		ProvenLeaves: proof.ProvenLeaves,
+		ProofNodes:   proof.ProofNodes,
+	}}
+
+	return out, nil
+}
+
+func (r *rpcServer) Execute(ctx context.Context, in *api.ExecuteRequest) (*api.ExecuteResponse, error) {
```
zalmen: Execute should be non-blocking. It should return once it validated the input.

proving/proving.go
```diff
@@ -15,63 +14,65 @@ const (
...
 		log.Error(err.Error())
 	}
 	return proof, err
 }
 
-func generateProof(id []byte, challenge Challenge, numOfProvenLabels uint8, difficulty Difficulty) (proof Proof,
-	err error) {
-
-	err = difficulty.Validate()
+func generateProof(id []byte, challenge Challenge, numOfProvenLabels uint8, difficulty Difficulty, dir string) (*Proof, error) {
```
zalmen: `generateProof` should validate that the initialization had completed before generating new proofs.

rpc/api/api.proto
moshababo: I thought it should be done by consuming the package directly, and not via the RPC server (which is basically the prover side). But we can add it to the RPC to easily support non-Go clients. 

And this is how we're testing now - proving via RPC server, and validating by using the package directly.

cmd/server/server.go
```diff
@@ -0,0 +1,150 @@
...
+
+var Cmd = &cobra.Command{
+	Use:   "server",
+	Short: "start server",
+	Run: func(cmd *cobra.Command, args []string) {
+		s := NewPostServer()
+
+		logger, err := s.Initialize(cmd, args)
+		if err != nil {
+			panic(err)
```
y0sher: maybe we can do a shutdown with cleanup instead of panicking? doesn't have to be as part of this pr

moshababo: There's no need for a cleanup yet, but I changed the panic into log and quit instead.

zalmen (approved):

PR #16, opened 2019-06-13
Benchmark util

moshababo:
Basic benchmark util, as a separate executable.


cmd/bench/bench.go
```diff
@@ -0,0 +1,157 @@
...
+			panic(err)
+		}
+		eInit := time.Since(t)
+
+		t = time.Now()
+		err = validator.Validate(proof)
+		if err != nil {
+			panic(err)
+		}
+		eInitV := time.Since(t)
```
noamnelke, line 87: Since the validation is very short (sub ms), consider running it more than once? (e.g. loop 1000 times and divide the total time by 1000)

Also consider not rounding it to the ms so it's possible to notice differences in validation time.

Same for the other validation bench.

cmd/bench/bench.go
```diff
@@ -0,0 +1,157 @@
...
+
+func main() {
+	datadir := flag.String("datadir", defaultConfig.DataDir, "filesystem datadir path")
+	space := flag.Uint64("space", 1<<23, "space per unit, in bytes")
+	single := flag.Bool("single", false, "whether to execute a single test instead of the complete set")
+	flag.Parse()
+
+	log.Printf("bench config: datadir: %v, space: %v", *datadir, bytefmt.ByteSize(*space))
+
+	cases := genTestCases(*datadir, *space, *single)
```
noamnelke: Consider allowing users to provide the cases manually. This would allow testing different specific cases quickly without changing code.

Pro-level would be to expose a way to generate a file with the test cases that `genTestCases` generates that the user can tweak and then run.

After typing this comment I almost didn't post it because I'm not sure it's worth the effort, but decided add this disclaimer so you can decide for yourself.

noamnelke (approved):

cmd/bench/bench.go
moshababo: I think that if result is below 1ms (and due to the rounding, if it's below 0.5ms), the result is meaningless anyway. 

But in order to avoid seeing `0ms` in the report, i'll round it up to a Microsecond.

What do you say?

cmd/bench/bench.go
moshababo: For now I thought to have 3 modes: single, mid, full.
Single - already supported, but need to add more flags to fully control the config.
Full - some config args are auto-generated, using `genTestCases`.
Mid - same as Full, but only the critical tests.

Some flags will be redundant when not in `single` mode. 

What do you say?

cmd/bench/bench.go
noamnelke: I'm trying to see how validation scales: we'll validate PoST proofs for all miners once per epoch, so let's say 5k miners. If it takes 1ms then: 1ms * 5k = 5s.

I agree that 5 seconds once per epoch isn't terrible, so I'm ok with only changing the rounding.

cmd/bench/bench.go
noamnelke: If you're gonna work on this I think my approach can be more flexible to use and easier to code.

Just expose `genTestCases` to the outside and make it return some `JSON` (or similar human-readable format). Then make the benchmark accept this same format. Doing so will allow users to manually control what runs and making changes to `genTestCases` will also be easy when it's separately exposed, imo.

But... up to you... this is just how I imaging I'd like it to work, as a user.

moshababo: I wanted to avoid the need for a smart user, and stick with simple the cmd args. 
So if the user wants control, he can use the `single` mode, and run it many times.
Otherwise, `genTestCases` will create the same important tests that we can benchmark periodically.

But i’ll keep that in mind and add it if needed.

PR #17, opened 2019-06-23
init: workers channel batches

moshababo:
CPU workers to produce label groups in batches, in attempt to reduce channel ops overhead.

### INIT 2GB | 6 in-file parallelism
 before: 2m7s
 after: 1m58s

time % before | time % after | function 
-- | -- | -- 
24 | 39 | github.com/spacemeshos/sha256-simd.blockAvx2 (cpu)
6 | 8 | syscall.Syscall (disk write)
14 | 12 | runtime.pthread_cond_signal (chan ops)
26 | 16 | runtime.pthread_cond_wait (chan ops)
8 | 1 | runtime.usleep 

### INIT 2GB | no parallelism
before: 10m7s
after: 7m43s
**no channels: 8m32s**

time % before | time % after | time % no-channels | function 
-- | -- | -- | --
0.5 | 6 | 11 |  github.com/spacemeshos/sha256-simd.blockAvx2 (cpu)
3 | 5 | 75 | syscall.Syscall (disk write)
59 | 73 | 2 | runtime.pthread_cond_signal (chan ops)
23 | 9 | 2 | runtime.pthread_cond_wait (chan ops)
10 | 1 | 0 | runtime.usleep

### DIFF
--- a/initialization/initialize.go
+++ b/initialization/initialize.go
@@ -199,42 +199,73 @@ func (init *Initializer) initFile(id []byte, fileIndex int, labelGroupsPerFile u
 		fileIndex, infileParallelism)
 
 	numOfWorkers := infileParallelism
-	workersOutputChans := make([]chan []byte, numOfWorkers)
+	workersOutputChans := make([]chan [][]byte, numOfWorkers)
 	errChan := make(chan error, 0)
 	finishedChan := make(chan struct{}, 0)
+	batchSize := 100
+	chanBuffer := 100
 
+	// CPU workers.
 	fileOffset := uint64(fileIndex) * labelGroupsPerFile
 	for i := 0; i < numOfWorkers; i++ {
 		i := i
-		workersOutputChans[i] = make(chan []byte, 1000) // TODO: make chan buffer configurable
-		offset := i
+		workersOutputChans[i] = make(chan [][]byte, chanBuffer)
+		workerOffset := i
 		go func() {
-			// Calculate labels in groups, write them to disk
-			// and append them as leaves in the merkle tree.
-			for position := uint64(offset); position < labelGroupsPerFile; position += uint64(numOfWorkers) {
-				lg := CalcLabelGroup(id, position+fileOffset, Difficulty(init.cfg.Difficulty))
-				workersOutputChans[i] <- lg
+			// Calculate labels in groups and write them to channel in batches.
+			iterator := 0
+			position := uint64(workerOffset)
+			batch := make([][]byte, batchSize)
+			for position < labelGroupsPerFile {
+				batch[iterator%batchSize] = CalcLabelGroup(id, position+fileOffset, Difficulty(init.cfg.Difficulty))
+
+				if iterator%batchSize == batchSize-1 {
+					workersOutputChans[i] <- batch
+					batch = make([][]byte, batchSize)
+				}
+
+				iterator += 1
+				position += uint64(numOfWorkers)
 			}
+			workersOutputChans[i] <- batch
 		}()
 	}
 
+	// IO worker.
 	go func() {
-		for i := uint64(0); i < labelGroupsPerFile; i++ {
-			// Consume workers output in a round-robin fashion.
-			lg := <-workersOutputChans[i%uint64(numOfWorkers)]
-			err := labelsWriter.Write(lg)
-			if err != nil {
-				errChan <- err
-				return
-			}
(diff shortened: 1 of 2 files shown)


initialization/initialize.go
```diff
@@ -199,42 +199,73 @@ func (init *Initializer) initFile(id []byte, fileIndex int, labelGroupsPerFile u
...
 			}
 
-			if (i+1)%init.cfg.LabelsLogRate == 0 {
-				init.logger.Info("initialization: file %v completed %v labels", fileIndex, i+1)
+			// Consume label groups from the batches in round-robin fashion.
+			for i := 0; i < batchSize*numOfWorkers; i++ {
+				batch := batches[i%numOfWorkers]
+				lg := batch[i/numOfWorkers]
+				if lg == nil {
+					break loop
```
zalmen: Why are you breaking the loop in such a case? If the batch size isn't a divider of the total number of labels you'll have batches that aren't full.

initialization/initialize.go
```diff
@@ -199,42 +199,73 @@ func (init *Initializer) initFile(id []byte, fileIndex int, labelGroupsPerFile u
...
-			// Calculate labels in groups, write them to disk
-			// and append them as leaves in the merkle tree.
-			for position := uint64(offset); position < labelGroupsPerFile; position += uint64(numOfWorkers) {
-				lg := CalcLabelGroup(id, position+fileOffset, Difficulty(init.cfg.Difficulty))
-				workersOutputChans[i] <- lg
+			// Calculate labels in groups and write them to channel in batches.
+			iterator := 0
+			position := uint64(workerOffset)
+			batch := make([][]byte, batchSize)
+			for position < labelGroupsPerFile {
```
zalmen, line 218: I think it is better if the batches will contain consecutive labels instead of labels in a round robin manner. It will simplify the IO part a lot

initialization/initialize.go
```diff
@@ -199,42 +199,73 @@ func (init *Initializer) initFile(id []byte, fileIndex int, labelGroupsPerFile u
...
-			err = tree.AddLeaf(lg)
-			if err != nil {
-				errChan <- err
-				return
+	loop:
+		for batchesIterator := 0; ; batchesIterator++ {
+			// Consume next batch from all workers.
+			batches := make([][][]byte, numOfWorkers)
+			for i, workerChan := range workersOutputChans {
+				batches[i] = <-workerChan
```
zalmen: If the batches contained consecutive labels (like it was suggested in my comment above) you wouldn't have had to wait for all routines to complete a round before starting to write to the disk.

zalmen (changes requested):

initialization/initialize.go
moshababo: Yes, but the first non-full batch indicates that all consecutive batches will be non-full as well, from the current index henceforth. So if i'm encountering a nil item, there are no more items.

initialization/initialize.go
moshababo: I originally started with this approach but it seemed to me even more complicated. I'll give it another try.

initialization/initialize.go
```diff
@@ -203,50 +203,42 @@ func (init *Initializer) initFile(id []byte, fileIndex int, labelGroupsPerFile u
...
 			batch := make([][]byte, batchSize)
 			for position < labelGroupsPerFile {
-				batch[iterator%batchSize] = CalcLabelGroup(id, position+fileOffset, Difficulty(init.cfg.Difficulty))
+				batch[position%uint64(batchSize)] = CalcLabelGroup(id, position+fileOffset, Difficulty(init.cfg.Difficulty))
 
-				if iterator%batchSize == batchSize-1 {
+				if position%uint64(batchSize) == uint64(batchSize-1) {
 					workersOutputChans[i] <- batch
 					batch = make([][]byte, batchSize)
+					position += uint64(numOfWorkers * batchSize)
```
zalmen: you're adding too much, you should add `uint64((numOfWorkers-1)*batchSize)`. How come there is no UT that failed on this? As I understand the code is broken this way.... Please add UT and make sure that it fails and then fix it

zalmen (changes requested):

initialization/initialize.go
moshababo: Good catch. The necessary UT fix was on a follow-up branch. Will add it now.

zalmen (approved):

PR #18, opened 2019-06-26
Execution phase read parallelism

moshababo:
Adding read parallelism support for the execution phase. 

## Benchmarks

SPACE | DISK | READ-PARALLELISM | EXEC | 
-- | -- | -- | -- | 
8GB | SSD | 1 | 134s
8GB | SSD | 6 | 33s
8GB | HDD | 1 | 134s
8GB | HDD | 6 | 30s


proving/proving.go
```diff
@@ -56,26 +60,68 @@ func (p *Prover) generateProof(id []byte, challenge Challenge) (*Proof, error) {
...
+	readers, err := persistence.GetReaders(dir)
 	if err != nil {
 		return nil, err
 	}
-	if width*difficulty.LabelsPerGroup() >= math.MaxUint64 {
-		return nil, fmt.Errorf("leaf reader too big, number of label groups (%d) * labels per group (%d) "+
-			"overflows uint64", width, difficulty.LabelsPerGroup())
+
+	parallelism := p.CalcParallelism(len(readers))
+	if parallelism > 1 {
```
noamnelke: I wonder if the existing method for "no parallelism" is more efficient than running the new method with a single worker.

I have a suspicion that the new method is actually more performant, but even if it's about the same -- removing this `if` and the old method would simplify the code and reduce the possibility for hidden bugs (better to have a single code path for a certain task).

I suggest benchmarking this to decide.

noamnelke (changes requested):
The changes look good in general, but I think we should consider using the same code path for single and multiple workers (since the multiple workers code path should work for a single worker as well).

proving/proving.go
moshababo: Good point.. 
The new method with a single worker is indeed a bit more performant. So there's no reason to maintain two different code paths.

noamnelke (approved):

PR #22, opened 2019-07-16
Crash recovery support

moshababo:
Closing https://github.com/spacemeshos/go-spacemesh/issues/737, https://github.com/spacemeshos/go-spacemesh/issues/753.

* If the initialization procedure crashes, that state would be detected in a consecutive procedure. The merkle tree state would be re-constructed from scratch, and the missing label groups (PoW) would be added.
* The main test for this functionality is `TestHarness_CrashRecovery`. It's using the integration harness because it's not straightforward to simulate a crash when not running on a separate process.
* I added space availability checks before performing the initialization, but they don't seem to be reliable enough. In the meanwhile, I added config option to disable them. If I won't find another reliable implementation, i'll remove it entirely. 
* RPC server protects against an attempt to perform multiple initializations per id, in parallel. The `initialization` package itself does not protect against that.


config/config.go
```diff
@@ -13,7 +13,7 @@ const (
 	// In bytes. 1 peta-byte of storage.
 	// This would protect against number of labels uint64 overflow as well,
 	// since the number of labels per byte can be 8 at most (3 extra bit shifts).
-	MaxSpace = 1 << 40 // 1099511627777
+	MaxSpace = 1 << 50 // 1099511627777
```
noamnelke: You're right that `1 << 50` is actually 1 petabyte, but the number in the comment doesn't match now. Maybe just remove it.

initialization/initializer.go
```diff
@@ -0,0 +1,547 @@
...
+
+func (init *Initializer) SetLogger(logger Logger) {
+	init.logger = logger
+}
+
+// Initialize takes an id (public key), space (in bytes), numProvenLabels and difficulty.
+// Difficulty determines the number of bits per label that are stored. Each leaf in the tree is 32 bytes = 256 bits.
+// The number of bits per label is 256 / LabelsPerGroup. LabelsPerGroup = 1 << difficulty.
+// Supported values range from 5 (8 bits per label) to 8 (1 bit per label).
+func (init *Initializer) Initialize() (*proving.Proof, error) {
```
noamnelke, line 94: The method comment should be updated to reflect the new design.

initialization/initializer.go
```diff
@@ -0,0 +1,547 @@
...
+
+	return proof, err
+}
+
+func (init *Initializer) Reset() error {
+	if err := init.VerifyStarted(); err != nil {
+		return err
+	}
+
+	err := os.RemoveAll(init.dir)
```
noamnelke: This will also remove the miner identity file (`key.bin`). We should be specific about the files we delete and remove the directory only if it's empty when we're done.

initialization/initializer.go
```diff
@@ -0,0 +1,547 @@
...
+		i := i
+		workersChans[i] = make(chan [][]byte, chanBufferSize)
+		workerOffset := i
+		go func() {
+			// Calculate labels in groups and write them to channel in batches.
+			position := uint64(workerOffset*batchSize) + existingWidth
+			batch := make([][]byte, batchSize)
+			batchPosition := uint64(0)
+
+			for batchPosition+position < labelGroupsPerFile {
```
noamnelke, line 365: I think it would really help readability if this section had some clarifying comments. Wasn't sure where to place this comment - it's relevant to the whole section incl. the external `for` loop.

integration/server.go
```diff
@@ -27,7 +28,8 @@ func newConfig(cfg *config.Config) (*serverConfig, error) {
 		return nil, err
 	}
 
-	postPath, err := postExecutablePath(baseDir)
+	postPath, err := postExecutablePath(filepath.Join(os.TempDir(), "post-build"))
```
noamnelke, line 31: I'm a little worried that we'll experience issues in the app test, where we run multiple nodes on the same machine / filesystem. How hard would it be to include the node ID in the path?

noamnelke (changes requested):
You should run `go mod tidy`

initialization/initializer.go
moshababo: Why wouldn't we want to delete the folder with all its files, including `key.bin`?
I think it leaves room for less errors.

Yes, if `Reset` is called, and then `Initialize`, the `key.bin` file won't be added again. But that's because it's not part of the `Initializer` functionality, but rather the init-factory cmd util. We can consider supporting it internally instead.

integration/server.go
moshababo: I added this to be able to persist the datadir between different runs, which previously contained also the build output. 
So it should not make things worse for what you described. 
If you already experienced problems with it, let me know.

moshababo: The point is to do compilation only once, so that doesn’t work with associating it with a specific id. 
I added this to be able to persist the datadir between different runs, which previously contained also the build output.
So it should not make things worse for what you described.

initialization/initializer.go
noamnelke: Hmm... On the one hand -- you make a good point. On the other -- while `key.bin` can be created by the init factory, it can now also be created by the node.

I'm not entirely sure when `Reset` is supposed to be called, other than the init factory. If the node calls it and knows to re-create the `key.bin` file it's fine, otherwise we may have an issue.

This dilemma makes me think that using the PoST data directory for the node's key file may have been a bad choice. In any case, we can't just ignore the issue.

We should either clean-up selectively in the PoST or fix it on the node side. Otherwise, if `Reset` shouldn't be called outside of the init factory, then we should remove it from the RPC server.

integration/server.go
noamnelke: ok, I just realized this is actually for the build and not for runtime data. 👍🏻

noamnelke (approved):

noamnelke:
We only need to decide what to do with the `Reset()` / `key.bin` issue. Either fix it now or create an issue for it.

moshababo:
I ended up making `Reset()` to remove files selectively.

PR #34, opened 2019-12-25
Less allocations in GetSha256Parent

### LINKED ISSUE #1617: Investigate Merkle Tree over memory allocation
Specifically in `AddLeaf`

noamnelke:
closes https://github.com/spacemeshos/go-spacemesh/issues/1617

---

@y0sher noticed that `GetSha256Parent` allocates a lot of memory during PoST proofs. Looking at the code, this makes sense: each `append` allocates memory for a new backing array.

I've looked into how I can reduce allocations and tried the following benchmarks:

```golang
func BenchmarkCombine(b *testing.B) {
	x := []byte("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
	y := []byte("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb")
	z := []byte("cccccccccccccccccccccccccccccccc")
	var q []byte

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		q = combineGlobal(x, y, z)
	}
	println(q)
}

func combineAppend(x, y, z []byte) []byte {
	message := append(x, append(y, z...)...)
	return message
	// 84.9 ns/op
}

func combineString(x, y, z []byte) []byte {
	message := []byte(string(x) + string(y) + string(z))
	return message
	// 112 ns/op
}

func combineCopy(x, y, z []byte) []byte {
	message := make([]byte, len(x) + len(y) + len(z))
	copy(message, x)
	copy(message[len(x):], y)
	copy(message[len(x)+len(y):], z)
	return message
	// 65.2 ns/op
}

var m []byte

func combineGlobal(x, y, z []byte) []byte {
	if len(m) != len(x) + len(y) + len(z) {
		m = make([]byte, len(x) + len(y) + len(z))
	}
	copy(m, x)
	copy(m[len(x):], y)
	copy(m[len(x)+len(y):], z)
	return m
	// 13.0 ns/op
}

var buf = bytes.Buffer{}

func combineBuff(x, y, z []byte) []byte {
	buf.Reset()
	buf.Write(x)
	buf.Write(y)
	buf.Write(z)
	return buf.Bytes()
	// 24.5 ns/op
}
```

`combineAppend` was what we previously did, and it looks like `combineString` is even worse (and probably causes more allocations, since each string has to be allocated separately), but it was a nice try.

`combineCopy` is an improvement, but it only really saves us a single allocation.

`combineGlobal` is what I ended up going with. It has a major downside that it's not thread safe, but it ammortizes the only allocation that doesn't have to happen, as long as the lengths of the byte arrays remain constant (in our case they do).

`combineBuff` could be a nice alternative. It's also not thread-safe, but it responds better to changing sizes and it's a little more readable, for a slightly increased runtime.

### DIFF
--- a/shared/challenge.go
+++ b/shared/challenge.go
@@ -6,17 +6,18 @@ type Challenge []byte
 
 var (
 	ZeroChallenge = make(Challenge, 0)
-	message []byte
+	buffer        []byte
 )
 
-// ⚠️ This method is NOT thread-safe
+// ⚠️ This method is NOT thread-safe. The code is optimized for performance and memory allocations.
 func (ch Challenge) GetSha256Parent(lChild, rChild []byte) []byte {
-	if len(message) != len(ch)+len(lChild)+len(rChild) {
-		message = make([]byte, len(ch)+len(lChild)+len(rChild))
+	l := len(ch) + len(lChild) + len(rChild)
+	if len(buffer) < l {
+		buffer = make([]byte, l)
 	}
-	copy(message, ch)
-	copy(message[len(ch):], lChild)
-	copy(message[len(ch)+len(lChild):], rChild)
-	res := sha256.Sum256(message)
+	copy(buffer, ch)
+	copy(buffer[len(ch):], lChild)
+	copy(buffer[len(ch)+len(lChild):], rChild)
+	res := sha256.Sum256(buffer[:l])
 	return res[:]
 }
--- a/shared/challenge.go
+++ b/shared/challenge.go
@@ -4,10 +4,19 @@ import "github.com/spacemeshos/sha256-simd"
 
 type Challenge []byte
 
-var ZeroChallenge = make(Challenge, 0)
+var (
+	ZeroChallenge = make(Challenge, 0)
+	message []byte
+)
 
+// ⚠️ This method is NOT thread-safe
 func (ch Challenge) GetSha256Parent(lChild, rChild []byte) []byte {
-	children := append(lChild, rChild...)
-	res := sha256.Sum256(append(ch, children...))
+	if len(message) != len(ch)+len(lChild)+len(rChild) {
+		message = make([]byte, len(ch)+len(lChild)+len(rChild))
+	}
+	copy(message, ch)
+	copy(message[len(ch):], lChild)
+	copy(message[len(ch)+len(lChild):], rChild)
+	res := sha256.Sum256(message)
 	return res[:]
 }


shared/challenge.go
```diff
@@ -4,10 +4,19 @@ import "github.com/spacemeshos/sha256-simd"
 
 type Challenge []byte
 
-var ZeroChallenge = make(Challenge, 0)
+var (
+	ZeroChallenge = make(Challenge, 0)
+	message []byte
```
y0sher: is it a message? maybe a better name that will indicate why we added this. or a name and a comment

y0sher:
I added one comment, also your benchmark mention performance (`// 13.0 ns/op`) but nothing about allocations (you should have `allocs` in the benchmark..) which is what we tried to reduce here..

shared/challenge.go
moshababo: I'd rather call it `buffer`, to explain its functionality.

shared/challenge.go
```diff
@@ -4,10 +4,19 @@ import "github.com/spacemeshos/sha256-simd"
...
+)
 
+// ⚠️ This method is NOT thread-safe
 func (ch Challenge) GetSha256Parent(lChild, rChild []byte) []byte {
-	children := append(lChild, rChild...)
-	res := sha256.Sum256(append(ch, children...))
```
moshababo, line 11: Perhaps add a comment here explaining that this is an optimized code.

shared/challenge.go
```diff
@@ -4,10 +4,19 @@ import "github.com/spacemeshos/sha256-simd"
...
+	if len(message) != len(ch)+len(lChild)+len(rChild) {
```
moshababo: We can work with `message` even if it's larger, no?
But since we don't expect this use-case here, it's not a must.

moshababo:
What about adding the benchmarks to the codebase?

y0sher:
> What about adding the benchmarks to the codebase?

+1

shared/challenge.go
noamnelke: Not bad! I've benchmarked the following and it's slightly faster:
```golang
func combineGlobal(x, y, z []byte) []byte {
	l := len(x) + len(y) + len(z)
	if len(m) < l {
		m = make([]byte, l)
	}
	copy(m, x)
	copy(m[len(x):], y)
	copy(m[len(x)+len(y):], z)
	return m[:l]
	// 11.7 ns/op
}
```

I'm pretty sure the compiler takes care of that `l` and it doesn't require an allocation. Will make sure before merging. If I use a global for `l` it's actually a little slower.

noamnelke:
I'm not sure about adding the benchmarks to the codebase. Maybe I'll add a benchmark for the final implementation, but the rest is research and I think it's place is in documentation, like this PR.

noamnelke:
I've run it with `pprof` and the allocations are just as I expected:
```golang
func BenchmarkCombine(b *testing.B) {
	x := []byte("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
	y := []byte("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb")
	z := []byte("cccccccccccccccccccccccccccccccc")
	var q []byte

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		q = combineGlobal(x, y, z)
	}
	println(q)
}

func combineAppend(x, y, z []byte) []byte {
	message := append(x, append(y, z...)...)
	return message
	// 84.9 ns/op | 2 allocations (64 + 96)
}

func combineString(x, y, z []byte) []byte {
	message := []byte(string(x) + string(y) + string(z))
	return message
	// 112 ns/op | 1 allocation (96)
}

func combineCopy(x, y, z []byte) []byte {
	message := make([]byte, len(x)+len(y)+len(z))
	copy(message, x)
	copy(message[len(x):], y)
	copy(message[len(x)+len(y):], z)
	return message
	// 65.2 ns/op | 1 allocation (96)
}

var m []byte

func combineGlobal(x, y, z []byte) []byte {
	l := len(x) + len(y) + len(z)
	if len(m) < l {
		m = make([]byte, l)
	}
	copy(m, x)
	copy(m[len(x):], y)
	copy(m[len(x)+len(y):], z)
	return m[:l]
	// 11.7 ns/op | no allocations
}

var buf = bytes.Buffer{}

func combineBuff(x, y, z []byte) []byte {
	buf.Reset()
	buf.Write(x)
	buf.Write(y)
	buf.Write(z)
	return buf.Bytes()
	// 24.5 ns/op | no allocations
}
```

noamnelke:
@y0sher see my previous comment. I ran it with `pprof` now.

moshababo (approved):

y0sher (approved):

noamnelke:
Final version of the benchmark:

```golang
var (
	x = []byte("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
	y = []byte("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb")
	z = []byte("cccccccccccccccccccccccccccccccc")
)

func BenchmarkCombineAppend(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = combineAppend(x, y, z)
	}
}

func BenchmarkCombineString(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = combineString(x, y, z)
	}
}

func BenchmarkCombineCopy(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = combineCopy(x, y, z)
	}
}

func BenchmarkCombineGlobal(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = combineGlobal(x, y, z)
	}
}

func BenchmarkCombineBuff(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_ = combineBuff(x, y, z)
	}
}

func combineAppend(x, y, z []byte) []byte {
	message := append(x, append(y, z...)...)
	return message
}

func combineString(x, y, z []byte) []byte {
	message := []byte(string(x) + string(y) + string(z))
	return message
}

func combineCopy(x, y, z []byte) []byte {
	message := make([]byte, len(x)+len(y)+len(z))
	copy(message, x)
	copy(message[len(x):], y)
	copy(message[len(x)+len(y):], z)
	return message
}

var m []byte

func combineGlobal(x, y, z []byte) []byte {
	l := len(x) + len(y) + len(z)
	if len(m) < l {
		m = make([]byte, l)
	}
	copy(m, x)
	copy(m[len(x):], y)
	copy(m[len(x)+len(y):], z)
	return m[:l]
}

var buf = bytes.Buffer{}

func combineBuff(x, y, z []byte) []byte {
	buf.Reset()
	buf.Write(x)
	buf.Write(y)
	buf.Write(z)
	return buf.Bytes()
}

/*
$ go test -bench '^BenchmarkCombine' -benchmem                                  [ 2s813 | Dec 25 03:11PM ]
goos: darwin
goarch: amd64
pkg: github.com/spacemeshos/post/shared
BenchmarkCombineAppend-8   	20000000	        79.3 ns/op	     160 B/op	       2 allocs/op
BenchmarkCombineString-8   	20000000	       106 ns/op	     192 B/op	       2 allocs/op
BenchmarkCombineCopy-8     	30000000	        41.7 ns/op	      96 B/op	       1 allocs/op
BenchmarkCombineGlobal-8   	100000000	        12.7 ns/op	       0 B/op	       0 allocs/op
BenchmarkCombineBuff-8     	50000000	        24.4 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/spacemeshos/post/shared	8.234s
 */
```

PR #35, opened 2019-12-25
Generate get parent func

noamnelke:
### Motivation
The node's `AppTest` runs several nodes in parallel, which causes the `GetSha256Parent ` function with a global buffer to return unexpected results (it's not thread safe).

### Changes
This replaces the static `GetSha256Parent` with `GenerateGetParentFunc`. The new function allocates a new `buffer` and generates the same function as before, but using a separate `buffer`.


y0sher (approved):
looks good to me, please test go-spacemesh with this.

noamnelke:
Tested `go-spacemesh` using this branch of `post`, all passed.

PR #38, opened 2020-06-29
Replace `ricochet2200/go-disk-usage` for FreeBSD compatibility

tzdybal:
`ricochet2200/go-disk-usage` is old, unmaintained library. It uses
deprecated `syscall` package, and it's incompatible with FreeBSD.

I created a fork, that solves both mentioned issues, is properly
versioned, adds Go module support, and I'm willing to maintain it,
if necessary.

This PR makes it possible to compile and run `post` on FreeBSD.
It's also the only code-level change required to use `go-spacemesh`
on FreeBSD.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/olekukonko/tablewriter v0.0.1
 	github.com/onsi/ginkgo v1.8.0 // indirect
 	github.com/onsi/gomega v1.5.0 // indirect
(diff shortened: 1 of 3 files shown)


lrettig (approved):
The new module looks good to me. Tested on macos and Linux. Will let @moshababo make the final call/merge this.

moshababo:
Thanks @tzdybal! LGTM.

PR #40, opened 2020-07-18
Go bindings for gpu-post lib

moshababo:
Go bindings for [gpu-post](https://github.com/spacemeshos/gpu-post) lib.

In addition:
* Checked-in latest `develop` branch artifacts for macOS/Linux/Windows (commit: https://github.com/spacemeshos/gpu-post/commit/665aaff762303f5f87f1ad5ff191d6131eca14f7)
* Added the `fetch_artifacts` tool (currently as a separate cmd; it will get extracted out of the repo later on). It currently supports fetching build artifacts of the latest commit of a given branch.
* Added GitHub Actions CI workflow for (mainly) testing the bindings on macOS/Linux/Windows. The test VM has no GPU capabilities, so no actual GPU-device testing could be run there.

PR #41, opened 2020-08-19
Finalize GPU/Tree-free proving

moshababo:
Finalizing GPU lib integration and Tree-free proving (following initial work done by #39 and #40).

Main remarks:

* Following https://github.com/spacemeshos/go-spacemesh/issues/1134, generating a proof returns, in addition to the raw proof, the proof metadata containing all the params that it is expected to be syntactically verified upon. 
* Added `InitOpts`, separated from `Config`, which defines params needed solely for PoST initialization, while `Config`  defines the protocol params.
* `id` and `salt` must be 32 bytes, otherwise junk memory data, from beyond the array boundary, will be used by `gpu-post` (C style).
* Applied global mutex locking for serializing calls to `gpu-post`, which isn’t thread-safe.
* The init directory which was previously created within the configured datadir for every distinct id is no longer used. Instead, the configured datadir is used directly, expected to serve one id only (Product request).
* `WorkOracleOne` function was added for invoking a single `scrypt` hash (on the contrary to `WorkOracle` which invokes a batch of `scrypt` hashes), used in proof validation, via CPU. It might later be replaced by a native Go implementation, or by new function in `gpu-post` which will accept a list of position indices to be hashed, potentially via GPU.
* `gpu-post` Go bindings: `Makefile` is set to fetch and bind https://github.com/spacemeshos/gpu-post/releases/tag/v0.1.21. Pre-compiled binaries are not checked-in (see https://github.com/spacemeshos/post/pull/43).
* Dependency to https://github.com/spacemeshos/merkle-tree was removed, since the tree-free proving is no longer using Merkle tree. 
* Added dependency to [bitstream](https://github.com/spacemeshos/bitstream) which provides bit-granular access to `io.Reader`/`io.Writer`. It is used in 3 different places: labels reading from disk in proof generation, indices writing in proof generation, indices reading in proof verification.
* Added support for incremental increase/decrease of `NumUnits` (currently only when number of files is 1).
* The following utils were ~~disabled~~ removed: RPC server, benchmark util, init factory util. All of them are outdated, and should be upgraded when we'll be certain that they are needed.


avive:
It will be great if commented-out dead code such as bench.go, init.go, config.go, server.go, repcserver.go, post.go, post_test.go, signal.go and the proto api definitions can be removed from this PR - it is really hard to understand what's going on in the project with all of code in these files commented out. They can be easily be added again if/when needed from older commits. Also consider modifying project structure to reflect current functionality, for example, remove the whole cmd source folder - I think that all the code in it is commented out.

It will also be nice if we can have basic instructions regarding how to build, test and run the project in the readme.

go.mod
```diff
@@ -1,29 +1,19 @@
...
-	github.com/tzdybal/go-disk-usage v1.0.0
-	golang.org/x/net v0.0.0-20190522155817-f3200d17e092
-	google.golang.org/genproto v0.0.0-20180817151627-c66870c02cf8
-	google.golang.org/grpc v1.21.0
+	github.com/stretchr/testify v1.7.0
+	golang.org/x/net v0.0.0-20210428140749-89ef3d95e781 // indirect
+	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
+	google.golang.org/genproto v0.0.0-20200513103714-09dca8ec2884
+	google.golang.org/grpc v1.29.1
+	google.golang.org/protobuf v1.26.0
```
noamnelke: Redundant with [this](https://github.com/spacemeshos/post/pull/41/files?file-filters%5B%5D=.mod#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R4) (line 4).

config/config.go
```diff
@@ -4,124 +4,123 @@ import (
...
+		MinNumUnits:   DefaultMinNumUnits,
+		K1:            DefaultK1,
+		K2:            DefaultK2,
 	}
+}
+
+type InitOpts struct {
+	DataDir           string
+	NumUnits          uint
+	NumFiles          uint
```
noamnelke, line 65: Let's change this to `MaxBytesPerFile`.

noamnelke: Also, we should cap the max file size to what the volume supports.

initialization/initialization.go
```diff
@@ -0,0 +1,463 @@
...
+			return nil
+		}
+
+		init.logger.Info("initialization: continuing to write file #%v; current number of labels: %v, target number of labels: %v, start position: %v", fileIndex, numLabelsWritten, fileNumLabels, fileOffset)
+	} else {
+		init.logger.Info("initialization: starting to write file #%v; target number of labels: %v, start position: %v", fileIndex, fileNumLabels, fileOffset)
+	}
+
+	currentPosition := numLabelsWritten
+	outputChan := make(chan []byte, 1024)
```
noamnelke, line 259: Perhaps `1024` should be dependent on the batch size. In other words, we should control the memory footprint of `outputChan`.

oracle/oracle.go
```diff
@@ -3,20 +3,44 @@ package oracle
...
+
+func WorkOracleOne(cpuProviderID uint, id []byte, position uint64, bitsPerLabel uint32) []byte {
+	salt := make([]byte, 32) // TODO(moshababo): apply salt
+
+	res, err := gpu.ScryptPositions(cpuProviderID, id, salt, position, position, bitsPerLabel)
+	if err != nil {
+		panic(err)
+	}
+
+	return res.Output
```
noamnelke: 1. This should reuse `WorkOracle()` instead of reimplementing it.
2. `cpuProviderID` should be calculated rather than provided as an argument.

shared/granularity_specific_io.go
```diff
@@ -0,0 +1,89 @@
...
+	} else {
+		// Bit-granular reader is using bitstream as a wrapper for the underlying reader.
+		br := bitstream.NewReader(rd)
+		gsReader.ReadNext = func() ([]byte, error) {
+			return br.Read(itemBitSize)
+		}
+		gsReader.ReadNextUintBE = func() (uint64, error) {
+			return br.ReadUint64BE(int(itemBitSize))
+		}
+	}
```
noamnelke, line 44: Is this `if` justified? The gran-reader lib should do this optimization.

proving/proving.go
```diff
@@ -0,0 +1,303 @@
...
+		return nil, err
+	}
+	gsReader := shared.NewGranSpecificReader(reader, p.cfg.BitsPerLabel)
+
+	numWorkers := endNonce - startNonce + 1
+	var indices []byte
+
+	workersChans := make([]chan []byte, numWorkers)
+	for i := range workersChans {
+		workersChans[i] = make(chan []byte, 1000) // TODO(moshababo): use numLabels/2 size instead? need just enough buffer to circulate between the two routines
```
noamnelke: I think we can remove the `TODO`. I would possibly consider using a smaller number than 1000, but even with this number the maximal potential "wasted" work is very reasonable:

```
(num workers - 1) * 1000 computations + 1000 disk reads
```

proving/proving.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+	}
+
+	// Check for an error from the IO worker.
+	select {
+	case err := <-errChan:
+		p.logger.Debug("proving: error: %v", err)
+		return nil, err
+	default:
+	}
```
noamnelke, line 301: If an io error occurred, we don't necessarily want to return an error. Because the worker chans are buffered, one of the workers may have succeeded after the error occurred, and then we want to return the successful result.

proving/proving.go
```diff
@@ -0,0 +1,303 @@
...
+			Found:    fmt.Sprintf("%d", m.LabelsPerUnit),
+			DataDir:  p.datadir,
+		}
+	}
+
+	return nil
+}
+
+func (p *Prover) tryNonce(ctx context.Context, numLabels uint64, ch Challenge, nonce uint32, readerChan <-chan []byte, difficulty uint64) ([]byte, error) {
+	var bitsPerIndex = uint(shared.NumBits(numLabels))
```
noamnelke: Let's rename `NumBits()` to `MinBitsForRepresentation()`?

proving/proving.go
```diff
@@ -0,0 +1,303 @@
...
+		case label, more := <-readerChan:
+			if !more {
+				return nil, fmt.Errorf("exhausted all labels; tried: %v, passed: %v, needed: %v", index, passed, p.cfg.K2)
+			}
+
+			hash := FastOracle(ch, nonce, label)
+
+			// Convert the fast oracle output's leading 64 bits to a number,
+			// so that it could be used to perform math comparisons.
+			hashNum := binary.LittleEndian.Uint64(hash[:])
```
noamnelke: Let's extract this to a method and use it here and in the verifier.

noamnelke (changes requested):
These are the comments we discussed during our call - sorry for not submitting them until now...

config/config.go
moshababo: Will be addressed in a follow-up PR.

shared/granularity_specific_io.go
moshababo: As discussed, we would like to keep a specialized code path for these 2 cases at this layer, since the third-party lib optimization isn’t guaranteed.

proving/proving.go
moshababo: Will be addressed in a follow-up PR.

oracle/oracle.go
moshababo: fix: https://github.com/spacemeshos/post/pull/41/commits/9e9a213b4e157fbab9c5603006e248adad13625e

proving/proving.go
moshababo: fix: https://github.com/spacemeshos/post/pull/41/commits/11d88773a670f2622696c39dd107e8f0ed075a9f

proving/proving.go
moshababo: I've used `BinaryRepresentationMinBits` instead.
fix: https://github.com/spacemeshos/post/pull/41/commits/ed18ce242190a224a5967ad412c273d9cee26688

proving/proving.go
moshababo: fix: https://github.com/spacemeshos/post/pull/41/commits/d3383a532b44a6c5d911fee1a61d858aeab369e5

initialization/initialization.go
moshababo: Will be addressed in a follow-up PR.

noamnelke (approved):

PR #43, opened 2021-06-14
Use shared library from executable folder

sudachen:
# Motivation
Allow to use a shared library from the executable folder and find a library in build folder to build go-spacemesh right.

# Changes
Linux(Elf)/OSX(Mach-O) binaries uses `rpath` declaration to find dynamic libraries on load and dynamic linkage stage. There is `$ORIGIN` macro understandable by dynamic linker as path from where image is starting. So PR adds this sort declaration to CGO_LDFLAGS. 

To run tests it's required to point dynamic linked to the folder where shared libraries (libpgu-post.so) exists so it's required to add CGO_LDFLAGS environment variable when go test is executing.

Since GPU-POST binaries can be changed/upgraded independently it's add rule into make file for downloading and extract appropriate set of binaries.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -16,9 +16,9 @@ const (
 	// and file truncating is byte-granular regardless of `BitsPerLabel` value.
 	DefaultComputeBatchSize = 1 << 14
 
-	// 1MB per unit. Temporary value.
+	// 1KB per unit. Temporary value.
 	DefaultBitsPerLabel  = 8
-	DefaultLabelsPerUnit = 1 << 20
+	DefaultLabelsPerUnit = 1 << 10
 
 	DefaultMaxNumUnits = 10
 	DefaultMinNumUnits = 1
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -139,7 +139,7 @@ func (init *Initializer) Initialize() error {
 	fileNumLabels := numLabels / uint64(init.opts.NumFiles)
 
 	init.logger.Info("initialization: starting to write %v file(s); number of units: %v, number of labels per unit: %v, number of bits per label: %v, datadir: %v",
-		init.opts.NumFiles, init.opts.NumUnits, init.cfg.LabelsPerUnit, init.cfg.BitsPerLabel, init.cfg, init.opts.DataDir)
+		init.opts.NumFiles, init.opts.NumUnits, init.cfg.LabelsPerUnit, init.cfg.BitsPerLabel, init.opts.DataDir)
 
 	for i := 0; i < int(init.opts.NumFiles); i++ {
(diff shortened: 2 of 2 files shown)


avive:
What are the steps you use to test this PR so it can be reviewed?

.github/workflows/ci.yml
```diff
@@ -17,4 +17,4 @@ jobs:
         with:
           go-version: '1.13.12'
```
avive: Why older version of go?

Makefile.Inc
```diff
@@ -0,0 +1,59 @@
...
+  endif
+endif
+
+$(info "OS: $(OS), HOST_OS: $(HOST_OS), GOOS: $(GOOS), GOARCH: $(GOARH), BIN_DIR: $(BIN_DIR), platform: $(platform)")
+
+GPU_SETUP_REV = 0.1.20
+GPU_SETUP_ZIP = libgpu-setup-$(platform)-$(GPU_SETUP_REV).zip
+GPU_SETUP_URL_ZIP = https://github.com/spacemeshos/gpu-post/releases/download/v$(GPU_SETUP_REV)/$(GPU_SETUP_ZIP)
+ifeq ($(platform), windows)
+  GPU_SETUP_LIBS = gpu-setup.dll
```
avive: Aren't we missing gpu-setup.lib here?

avive (changes requested):

Makefile.Inc
sudachen: Because gpu-setup.lib is build by VS and in MSCOFF format, but go uses GNU toolchain and requires DWARF. Icreated required library and you can add to zip. It's here https://github.com/sudachen/post/blob/playground/gpu/libgpu-setup-win64.a

.github/workflows/ci.yml
sudachen: Because I did not change it. I prefer to use 1.16.5.

Makefile.Inc
sudachen: No. It's not usable by go because it uses gnu toolchain, but gpu-setup.lib is built by VS and in the MSCOFF format.

sudachen:
> What are the steps you use to test this PR so it can be reviewed?

It's tested with CI on ubuntu/windows/macos automatically on PR update with go test.

.github/workflows/ci.yml
avive: I think we should use 1.16 but we need to make sure this will be compatible with go-spacemesh build min version.

Makefile.Inc
avive: Good to know - thanks for the info.

PR #45, opened 2021-07-29
Fix data races

nkryuchkov:
There were multiple data races when running tests with the `-race` flag. This PR fixes them, also formats files with `gofumpt`, and fixes some issues found by the linter.


initialization/initialization.go
```diff
@@ -80,18 +81,18 @@ func NewInitializer(cfg Config, opts config.InitOpts, id []byte) (*Initializer,
...
 // pseudo-random data with respect to a specific id. This data is the result of a computationally-expensive operation.
 func (init *Initializer) Initialize() error {
-	init.mtx.Lock()
-	if init.initializing {
-		init.mtx.Unlock()
+	if init.isInitializing() {
 		return ErrAlreadyInitializing
 	}
+
 	init.stopChan = make(chan struct{})
```
moshababo, line 92: By limiting the scope of the previously applied lock, multiple goroutines can now enter this code path, no?

initialization/initialization.go
```diff
@@ -342,12 +361,16 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
 }
 
 func (init *Initializer) updateSessionNumLabelsWritten(numLabelsWritten uint64) {
+	init.mtx.RLock()
+
 	init.numLabelsWritten = numLabelsWritten
```
moshababo: Any reason for keeping this inside of the new lock scope?

proving/proving.go
```diff
@@ -271,8 +275,18 @@ func (p *Prover) tryNonces(numLabels uint64, challenge Challenge, startNonce, en
 		i := i
 		go func() {
 			nonce := startNonce + i
-			indices, err = p.tryNonce(ctx, numLabels, challenge, nonce, workersChans[i], difficulty)
-			resultsChan <- nonceResult{nonce, indices, err}
+
```
moshababo: I'm not sure I understand why these new steps are needed.

moshababo (commented):
Thanks for looking into this!

initialization/initialization.go
nkryuchkov: No, it may be removed outside of the scope.

initialization/initialization.go
nkryuchkov: Yes, until `init.setInitializing(true)` where the next mutex lock is located

proving/proving.go
nkryuchkov: Accessing the `indices` variable outside the goroutine requires a mutex for both reading a writing. Otherwise, there's a data race. However, it may be moved inside the goroutine, so there's no need for mutexes.

nkryuchkov: Actually, I see that it was already removed by Dmitry in his PR https://github.com/spacemeshos/post/pull/52

initialization/initialization.go
moshababo: Then this is an issue to be fixed, even if quite an edge case.

moshababo (changes requested):

moshababo:
Side note regarding https://github.com/spacemeshos/post/pull/45#discussion_r700608046: you've caught there a bug later realized by @dshulyak in #52, which had quite a bad ripple effect.

initialization/initialization.go
nkryuchkov: Fixed

moshababo (approved):

moshababo:
Thanks, I'm merging despite tests failure on Windows (will be resolved by https://github.com/spacemeshos/post/pull/54).

PR #47, opened 2021-08-12
Fix leaking goroutines in tryNonces

### LINKED ISSUE #46: Leaking goroutines in tryNonces
Multiple miners in 0.2 devnet are leaking goroutines:

```
**goroutine profile: total 598
323 @ 0x43c745 0x407d0a 0x407ab5 0xbe1d87 0x472161
#	0xbe1d86	github.com/spacemeshos/post/proving.(*Prover).tryNonces.func1+0x66	/go/pkg/mod/github.com/spacemeshos/post@v0.0.0-20210726170806-5cf91a7ae467/proving/proving.go:263

30 @ 0x43c745 0x44c6cf 0xe222f4 0x472161
#	0xe222f3	github.com/spacemeshos/go-spacemesh/p2p/net.(*MsgConnection).sendListener+0xf3	/go/src/github.com/spacemeshos/go-spacemesh/p2p/net/msgcon.go:157

23 @ 0x43c745 0x44c6cf 0xe2ce05 0xe23bea 0x472161
#	0xe2ce04	github.com/spacemeshos/go-spacemesh/p2p/net.(*udpConnWrapper).Read+0xc4			/go/src/github.com/spacemeshos/go-spacemesh/p2p/net/udp.go:402
#	0xe23be9	github.com/spacemeshos/go-spacemesh/p2p/net.(*MsgConnection).beginEventProcessing+0x89	/go/src/github.com/spacemeshos/go-spacemesh/p2p/net/msgcon.go:265**
```

Stack frames are accurate for current head of develop (https://github.com/spacemeshos/post/commit/5cf91a7ae467476520b8a44bc3644646fcc5d829)

dshulyak:
closes: https://github.com/spacemeshos/post/issues/46

- if worker exited stop sending more work to the worker
- make work producer wait on context as well, so that if we exit early it will exit too


dshulyak:
@moshababo let's merge this change at the start of the week please. this is the last leak that we have in the node, i want to include it in the next rc

dshulyak:
this was tested on latest devnet

proving/proving.go
```diff
@@ -237,68 +240,81 @@ func (p *Prover) tryNonces(numLabels uint64, challenge Challenge, startNonce, en
...
+		select {
+		case result = <-resultsChan:
+			if result.err != nil {
+				p.logger.Debug("proving: nonce %v failed: %v", result.nonce, result.err)
+			} else {
+				p.logger.Debug("proving: nonce %v succeeded", result.nonce)
+				return result, nil
+			}
+		case err := <-errChan:
+			if err != nil && !errors.Is(err, io.EOF) {
```
moshababo: Does the `err != nil` check needed? It was already checked before.

proving/proving.go
```diff
@@ -237,68 +240,81 @@ func (p *Prover) tryNonces(numLabels uint64, challenge Challenge, startNonce, en
...
-	case err := <-errChan:
-		p.logger.Debug("proving: error: %v", err)
-		return nil, err
-	default:
+	if result == nil {
+		panic("unexpected exit with nil result")
 	}
-
-	return solutionNonceResult, nil
+	return nil, result.err
```
moshababo: This function shouldn't return an error of a specific/arbitrary `tryNonce` result, but only an IO error. 
In case of no successful nonce and no IO error, both returned values should be nil. 
Otherwise, the error will be propagated back to `GenerateProof` outcome, and no additional iteration will be made.

proving/proving.go
moshababo: Is there a rationale for not checking `!errors.Is(err, io.EOF)` before feeding the error to the channel?

proving/proving.go
```diff
@@ -237,68 +240,81 @@ func (p *Prover) tryNonces(numLabels uint64, challenge Challenge, startNonce, en
 	var indices []byte
 
 	workersChans := make([]chan []byte, numWorkers)
+	// workersComplete channel will be closed when worker stops listening for appropriate workersChan
+	workersComplete := make([]chan struct{}, numWorkers)
 	for i := range workersChans {
-		workersChans[i] = make(chan []byte, 1024)
+		// by having a buffer of 1 all goroutines will be scheduled in a fair way
+		workersChans[i] = make(chan []byte, 1)
```
moshababo, line 246: Wouldn't the 1 buffer size impose performance limitation, where the IO and the compute workers are completely bound by each other? 
Besides having a slightly better memory footprint, is there another reason to impose this?

proving/proving.go
```diff
@@ -237,68 +240,81 @@ func (p *Prover) tryNonces(numLabels uint64, challenge Challenge, startNonce, en
...
 	}
-	resultsChan := make(chan nonceResult, numWorkers)
-	errChan := make(chan error, numWorkers)
-
+	resultsChan := make(chan *nonceResult, numWorkers)
+	errChan := make(chan error, 1)
+	var wg sync.WaitGroup
+	defer wg.Wait()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
moshababo, line 255: Better to space out this code block a bit with an empty line or two.

moshababo (commented):
Thanks, looks good.

proving/proving.go
dshulyak: the idea is that workers are supposed to make progress at the same speed. if there are not enough CPU per worker then some will be executed more often than others, by having this buffer we prevent some workers starving for CPU.

tbh i don't think that there will be any performance difference at all, it just seemed cleaner to me. i can revert

proving/proving.go
dshulyak: i left it just for safety

dshulyak: no rationale, just seemed cleaner to have it next to return from the method

proving/proving.go
dshulyak: :+1:

dshulyak:
@moshababo thanks for review. i adjusted code based on the suggestions. ptal

proving/proving.go
moshababo: Thanks for clarifying. We can keep this for now, and do some benchmarking later.

proving/proving.go
```diff
@@ -237,68 +240,79 @@ func (p *Prover) tryNonces(numLabels uint64, challenge Challenge, startNonce, en
...
-			resultsChan <- nonceResult{nonce, indices, err}
+			close(workersComplete[i])
+			resultsChan <- &nonceResult{nonce, indices, err}
+			wg.Done()
 		}()
 	}
 
 	// Drain the workers results chan.
-	var solutionNonceResult *nonceResult
+	var result *nonceResult
```
moshababo: After your fix for https://github.com/spacemeshos/post/pull/47#discussion_r695368434, it doesn't look like we need this anymore.

moshababo (approved):
Feel free to merge after addressing this last comment.

PR #52, opened 2021-08-31
proving: remove shared indices variable

dshulyak:
This was leading to a race with severe consequences, results from different tryNonce workers may mix up. More in https://github.com/spacemeshos/go-spacemesh/issues/2735.

`var indices []byte` is a shared variable that can be updated by concurrent goroutines. Consider this interleaving:

| G1 | G2 | |
| ---     | --- | --- | 
| indices = tryNonce(11) | | G1 completes tryNonce first |
| | indices = tryNonce(99) | G2 completes second, but before result from G1 is sent to channel |
| resultsChan <- result{indices} | resutlsChan <- result{indices} | both G1 and G2 are using indices from G2 | 

I think my previous change made it more likely to happen since I am always returning first proof sent to the channel, and previously we were returning the last proof sent. But it was still possible, because we can't guarantee any particular order. So if we change interleaving example slightly we can see how it can fail if we consume always last result:

| G1 | G2 | |
| ---     | --- | --- | 
| | indices = tryNonce(99) | G2 completes tryNonce first |
| indices = tryNonce(11) | | G1 completes second, but before result from G2 is sent to channel |
| resultsChan <- result{indices} | resutlsChan <- result{indices} | both G1 and G2 are using indices from G1 |

### DIFF
--- a/proving/proving.go
+++ b/proving/proving.go
@@ -237,8 +237,6 @@ func (p *Prover) tryNonces(numLabels uint64, challenge Challenge, startNonce, en
 	gsReader := shared.NewGranSpecificReader(reader, p.cfg.BitsPerLabel)
 
 	numWorkers := endNonce - startNonce + 1
-	var indices []byte
-
 	workersChans := make([]chan []byte, numWorkers)
 	// workersComplete channel will be closed when worker stops listening for appropriate workersChan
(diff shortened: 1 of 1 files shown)


moshababo (approved):
Good catch, it's definitely a bug in my original version; there's no need for a shared variable. It's probably a leftover from a previous implementation version.

PR #56, opened 2022-07-04
show stat about generationg PoS files

abergasov:
related issue https://github.com/spacemeshos/go-spacemesh/issues/3281

for serving full status about generating PoS files was added new method.

as result - returning this struct
```
type InitializerStat struct {
	TotalFilesToWrite uint64    // total number of files to write
	CompletedFiles    uint64    // number of files that were actually written
	LabelsPerFile    LabelInfo{
	  Label        uint64
	  LabelInBytes uint64
        } // number of labels in each file
	WrittenAtLastFile LabelInfo // number of labels written to the last file
}
```

this allow us to show user such state

1. how much percentage is done, like `86% completed`
2. show how much storage was generated, lke `170gb of 256 gb was smeshed`

failed ci on win is related to github actions, not related to this pr. issue https://github.com/spacemeshos/post/issues/58
but probably it will be fixed by golang dev team

### DIFF
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -44,9 +44,12 @@ func CPUProviderID() int {
 }
 
 type Initializer struct {
+	// data per file
 	// numLabelsWritten should be aligned by 8 bytes because it's accessed by atomics.
 	numLabelsWritten     uint64
 	numLabelsWrittenChan chan uint64
+
+	processedFiles uint64
 
 	cfg  Config
 	opts InitOpts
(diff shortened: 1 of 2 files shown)


moshababo:
I'm not sure regarding the utility of this. You mentioned the following: 
> 1. how much percentage is done, like 86% completed
> 2. show how much storage was generated, lke 170gb of 256 gb was smeshed

Yet both can be achieved by the current
```golang
func (init *Initializer) SessionNumLabelsWrittenChan() <-chan uint64
```

Which returns a stream of updates for the global, cross-file position. 
Meanwhile, you have defined the following properties: 

* `TotalFilesToWrite`: value is set by the user, prior to the session. 
* `CompletedFiles`: current position returned from `SessionNumLabelsWrittenChan()` implies the number of completed files (number of labels can be easily calculated, and number of files is set by the user).
* `LabelsPerFile`:  can be easily calculated by the user (also, we might change the API to have it be explicitly defined by the user, instead of `NumFiles`).
* `WrittenAtLastFile`: relies on the same value as `SessionNumLabelsWrittenChan()`.

Anything I'm missing?

PR #60, opened 2022-08-02
Scale codec for proof - migrate uint to uint32/uint64

evt:


shared/proof.go
```diff
@@ -45,9 +47,9 @@ func (p *Proof) Decode(data []byte) error {
 type ProofMetadata struct {
 	ID            []byte
 	Challenge     Challenge
-	BitsPerLabel  uint
-	LabelsPerUnit uint
-	NumUnits      uint
-	K1            uint
-	K2            uint
+	BitsPerLabel  uint32
+	LabelsPerUnit uint32
```
dshulyak, line 51: Noam commented yesterday that it should be uint64

PR #62, opened 2022-08-16
Fix build on arm64 architectures

fasmat:
This fixes the build on arm64, tested on Linux and macOS.

The upgrade to Go 1.19 was necessary to fix the complaints about missing memory during a test with `-race` on Windows


pigmej:
Checked on my mac, provided feedback and now it fully compiles and passes tests suite :)

proving/proving.go
```diff
@@ -150,7 +150,7 @@ func (p *Prover) loadMetadata() (*initialization.Metadata, error) {
 }
 
 func (p *Prover) verifyMetadata(m *Metadata) error {
-	if bytes.Compare(p.id, m.ID) != 0 {
+	if !bytes.Equal(p.id, m.ID) {
```
dshulyak, line 153: is it a part of the fix or function was changed in 1.19?

dshulyak (approved):

proving/proving.go
fasmat: golangci-lint complained on my machine, that `bytes.Equal` should be used over `bytes.Compare` when possible. That's why I changed it; functionally it's identical and both functions have been available since Go 1.0

PR #67, opened 2022-10-12
Amortise cost for nonce in POPS-VRF

### LINKED ISSUE #175: Amortize the cost of finding a nonce for POPS-VRF
## The task

Amortize the computation cost of finding a nonce to support the [POPS-VRF](https://github.com/spacemeshos/pm/issues/172) scheme through the cost of PoST initialization. Then apply it to the initial ATX.

## Background

https://github.com/spacemeshos/gpu-post supports the amortization via the `COMPUTE_POW` flag (which was added in addition to the `COMPUTE_LEAFS` flag, needed for the actual initialization). For more info, see [here](https://github.com/spacemeshos/SMIPS/issues/45).

## Breakdown 

https://github.com/spacemeshos/post & https://github.com/spacemeshos/go-spacemesh

* Enable the activation of the `COMPUTE_POW` flag, and use it in order to find a nonce which satisfies `H( nonce ++ VRFpk ++ PositioningATX ) < d`. 
* Create a calculation function for the difficulty threshold `d`. It should define a probability of finding one single nonce for the entire PoST initialization, meaning one for every (number of units) *  (number of labels per unit), when applied to the total range of values space in 32 bytes.
* If no successful nonce was found during the entire initialization phase, continue to compute labels for nonce-finding, PoW purposes only (`COMPUTE_LEAFS` off, `COMPUTE_POW` on), until succeeding.
* Upon finding a successful nonce
   * Add it to the initial ATX
   * Add it to the PoST metadata file
 * Verify incoming initial-ATXs' nonce to satisfy `H( nonce ++ VRFpk ++ PositioningATX ) < d`. The hash function is the work oracle function from PoST initialization, also used in PoST verification to compute a given label.

fasmat:
This PR adds looking for a nonce to satisfy a given difficulty during PoST initialization.

- The difficulty function used to find a nonce can be adapted to make it easier (or harder) to find a suitable nonce.
  - **TODO**: I believe the difficulty function as currently implemented is unfit for its purpose. See the comment I added to its godoc
- PostMetadata was updated to include the nonce and instead of the commitment both the node ID and the commitment ATX ID
  - the actual commitment is calculated by the `Initializer` from the node ID and the commitment ATX ID.
  - At the moment the used commitment ATX is stored in a nodes local database in `go-spacemesh`. This change will store it with the post data and simplify switching identities and offloading post generation to 3rd parties.
- A `VerifyPow` function was added that allows to verify if a given nonce satisfies the required difficulty.
- Introduced functional options for `WorkOracle` and `Initializer`
  - Parameter validation was moved into the functional options and the `validate()` method that is called when options are evaluated
- I simplified the `Initializer` not to use a worker routine and an I/O routine.
  - The `writer` used by the I/O worker is buffered and even with the fastest available GPUs writing the data to disk should take a fraction of the time it takes calculating it.
  - We can benchmark this in the future and reintroduce the worker & I/O routine split if we deem it beneficiary.


moshababo:
Is it ready for review? If so, what's the expected content? It doesn't contain yet the actual amortization.

fasmat:
@moshababo this is still a work in progress; that's why it's a Draft PR 🙂

gpu/api.h
```diff
@@ -25,6 +25,7 @@ extern "C" {
 #define	SPACEMESH_API_ERROR_CANCELED			-4
 #define	SPACEMESH_API_ERROR_NO_COMPOTE_OPTIONS	-5
 #define	SPACEMESH_API_ERROR_INVALID_PARAMETER	-6
+#define	SPACEMESH_API_ERROR_INVALID_PROVIDER_ID	-7
```
fasmat: I had to manually update this header by copying from the `gpu-post` release zip file. This should be done automatically by calling `make get-gpu-setup`

Makefile.Inc
```diff
@@ -55,12 +55,12 @@ GPU_SETUP_REV = 0.1.24
 GPU_SETUP_ZIP = libgpu-setup-$(platform)-$(GPU_SETUP_REV).zip
 GPU_SETUP_URL_ZIP = https://github.com/spacemeshos/gpu-post/releases/download/v$(GPU_SETUP_REV)/$(GPU_SETUP_ZIP)
 ifeq ($(platform), windows)
-  GPU_SETUP_LIBS = gpu-setup.dll
+  GPU_SETUP_LIBS = gpu-setup.dll gpu-setup.lib api.h
```
fasmat, line 58: Try renaming to `libgpu-setup.dll` or `libgpu-setup-win64.dll`

verifying/verifying.go
```diff
@@ -14,11 +15,37 @@ type (
...
 var (
-	WorkOracleOne = oracle.WorkOracleOne
-	FastOracle    = oracle.FastOracle
-	UInt64LE      = shared.UInt64LE
+	WorkOracle = oracle.WorkOracle
+	FastOracle = oracle.FastOracle
+	UInt64LE   = shared.UInt64LE
 )
 
+func VerifyPoW(nonce, numUnits, bitsPerLabel uint64, commitment []byte) error {
```
fasmat, line 23: Add tests and documentation for this function

initialization/initialization_test.go
```diff
@@ -55,9 +57,45 @@ func TestInitialize(t *testing.T) {
...
-
-		r.Equal(uint64(cfg.MinNumUnits)*cfg.LabelsPerUnit, init.SessionNumLabelsWritten())
 	}
+	r.Equal(uint64(cfg.MinNumUnits)*cfg.LabelsPerUnit, init.SessionNumLabelsWritten())
+
+	r.NotNil(init.nonce)
+	r.NoError(verifying.VerifyPoW(*init.nonce, uint64(cfg.MinNumUnits), uint64(cfg.BitsPerLabel), init.commitment))
+}
+
+func TestInitialize_PowOutOfRange(t *testing.T) {
```
fasmat, line 67: Add asserts to other tests to verify found nonce

Add test that stops initializing with and without nonce

oracle/oracle.go
```diff
@@ -3,44 +3,136 @@ package oracle
...
+
+		opts.difficulty = difficulty
+		return nil
+	}
+}
+
+type WorkOracleResult struct {
+	Output []byte
+	Nonce  *uint64
+}
```
fasmat, line 104: Add documentation

oracle/oracle.go
```diff
@@ -3,44 +3,136 @@ package oracle
...
+
+func WorkOracle(opts ...workOracleOptionFunc) (WorkOracleResult, error) {
```
fasmat, line 106: Add documentation

verifying/verifying_test.go
```diff
@@ -33,7 +33,7 @@ func getTestConfig(t *testing.T) (config.Config, config.InitOpts) {
 	opts.DataDir = t.TempDir()
 	opts.NumUnits = cfg.MinNumUnits
 	opts.NumFiles = 2
-	opts.ComputeProviderID = CPUProviderID()
+	opts.ComputeProviderID = int(CPUProviderID())
 
 	return cfg, opts
```
fasmat, line 38: Add tests for `VerifyPoW`

gpu/api.go
```diff
@@ -16,68 +16,185 @@ type ComputeProvider struct {
...
+	panic("no CPU provider available")
 }
 
+// Providers returns a list of available compute providers.
 func Providers() []ComputeProvider {
 	return providers
 }
 
-func CPUProviderID() int {
+// CPUProviderID returns the ID of the CPU provider or nil if the CPU provider is not available.
```
moshababo: "... or nil if the CPU provider is not available".
It doesn't seem like an accurate description.

initialization/initialization.go
```diff
@@ -44,61 +42,90 @@ var (
...
 }
 
-type initializeOption struct {
-	commitment []byte
-	cfg        *Config
-	initOpts   *config.InitOpts
-	logger     Logger
+type option struct {
+	nodeId []byte
+	atxId  []byte
```
moshababo: Calling it `commitmentAtxId` would be better. (same for functions related to this field)

initialization/initialization.go
```diff
@@ -170,14 +203,38 @@ func (init *Initializer) Initialize(ctx context.Context) error {
...
+	}
+
+	if init.nonce == nil {
+		// continue searching for a nonce
+		res, err := oracle.WorkOracle(
+			oracle.WithComputeProviderID(uint(init.opts.ComputeProviderID)),
+			oracle.WithNodeId(init.nodeId),
+			oracle.WithAtxId(init.atxId),
+			oracle.WithStartAndEndPosition(numLabels, 8*numLabels),
+			oracle.WithBitsPerLabel(uint32(init.cfg.BitsPerLabel)),
```
moshababo: Is this needed when `ComputeLeaves` is off?

initialization/initialization.go
```diff
@@ -170,14 +203,38 @@ func (init *Initializer) Initialize(ctx context.Context) error {
...
 	for i := 0; i < int(init.opts.NumFiles); i++ {
-		if err := init.initFile(ctx, uint(init.opts.ComputeProviderID), i, numLabels, fileNumLabels); err != nil {
+		if err := init.initFile(ctx, i, numLabels, fileNumLabels, difficulty); err != nil {
+			return err
+		}
+	}
+
+	if init.nonce == nil {
+		// continue searching for a nonce
+		res, err := oracle.WorkOracle(
```
moshababo, line 260: Was this functionality tested for recovery mode? (user doesn't find a nonce during initialization, continues here, and then restarts)

initialization/initialization.go
```diff
@@ -170,14 +203,38 @@ func (init *Initializer) Initialize(ctx context.Context) error {
...
+			oracle.WithComputeProviderID(uint(init.opts.ComputeProviderID)),
+			oracle.WithNodeId(init.nodeId),
+			oracle.WithAtxId(init.atxId),
+			oracle.WithStartAndEndPosition(numLabels, 8*numLabels),
```
moshababo: Why end at `8*numLabels` and not `math.MaxUint64`?
In addition to not providing full coverage, it can also lead to an overflow. 

Also, it might be better to split it into batches, at least for providing some feedback to the user.

initialization/initialization.go
```diff
@@ -404,11 +465,13 @@ func (init *Initializer) verifyMetadata(m *Metadata) error {
 
 func (init *Initializer) saveMetadata() error {
 	v := Metadata{
-		Commitment:    init.commitment,
+		NodeId:        init.nodeId,
+		AtxId:         init.atxId,
```
moshababo: Same here, it would be better to explicitly call it `commitmentAtxId`.

initialization/initialization.go
moshababo, line 260: Better to add logs before and after—this is a potentially long (and blocking) operation.

initialization/initialization.go
```diff
@@ -44,61 +42,90 @@ var (
...
+			return fmt.Errorf("invalid `id` length; expected: 32, given: %v", len(nodeId))
 		}
-		opts.commitment = commitment
+
+		opts.nodeId = nodeId
+		return nil
+	}
+}
+
+// WithAtxId sets the ID of the CommitmentATX.
```
moshababo: Not critical, but unlike `Initializer`, the work oracle doesn't need to know about `atxId` and `nodeId`, but rather only about `commitment`. This design also adds a few unnecessary hashes (once per batch invocation, instead of a single one in `Initializer`).

fasmat: Moved `CommitmentBytes` function to `oracle` package. `WorkOracle` now takes a `WithCommitment` functional option, while callers of the oracle use the `CommitmentBytes` function to calculate the commitment from the Node ID and Commitment ATX ID.

initialization/initialization.go
fasmat: I updated the code to check until `math.MaxUint64` in batches with the same size that is used to compute leaves. What do you suggest to provide feedback to the user? I don't consider logs as valuable feedback; they are for developers to debug and to include in bug reports.

While computing leaves the user feedback is the number of labels written, which is also used by Smapp to display progress.

The problem with how we are now looking for a nonce is that we don't know how long it will take. At least 1 in 50 semshers will have to check > 4x the number of labels as is required to compute leaves with no way to know in advance that it will take this long.

initialization/initialization.go
fasmat: Just adding logs isn't sufficient, see my comment here: https://github.com/spacemeshos/post/pull/67#discussion_r1033362119

fasmat: A user that finished initialization of their PoST, hasn't found a nonce, stops and restarts will have to start looking for a nonce at the `numLabels` index again.

The only way I see around this is to add the last position that was passed to WorkOracle to the metadata file. This would require a few additional checks (and tests for them) as well:

- what if `lastPosition` is less than `numLabels` and post file size doesn't match? - ignore `lastPosition` and continue initialization as it is implemented at the moment?
- what if `lastPosition` is greater than `numLabels` but post file size doesn't match? - restart?

initialization/initialization.go
fasmat: No it isn't; I updated the code accordingly.

initialization/initialization.go
moshababo: `initFile` already supports the initialization recovery by reading the number of written bytes on disk (in contrary to the metadata file, which is less reliable).
You only need to take care of the PoW-only mode recovery, by persisting `lastPosition` in the metadata file prior to every work oracle invocation at that stage (and not during initialization).

> - what if `lastPosition` is less than `numLabels` and post file size doesn't match? - ignore `lastPosition` and continue initialization as it is implemented at the moment?

This can happen if the user increased the number of units (or due to a bug or disk corruption). You should ignore `lastPosition` and continue initialization as-is. 

> - what if `lastPosition` is greater than `numLabels` but post file size doesn't match? - restart?

Same, if the user decreased the number of units. You should ignore `lastPosition` and continue initialization as-is.

initialization/initialization.go
moshababo: As with initialization, via logs + pub/sub channel. If we have batching+logs, the channel and Smapp support can be added later.

initialization/initialization.go
moshababo: Let's continue there.

fasmat: I added `LastPosition` to `Metadata`. It is set when no nonce is found below `numLabels` and if stopped and continued `Initializer` will start looking for nonces at higher positions.

initialization/initialization.go
fasmat: Added logs

initialization/initialization.go
```diff
@@ -170,17 +213,68 @@ func (init *Initializer) Initialize(ctx context.Context) error {
...
+		init.lastPosition = new(uint64)
+		*init.lastPosition = numLabels
+	}
+
+	// continue searching for a nonce
+	// TODO(mafa): PM-195 depending on the difficulty function this can take a VERY long time, with the current difficulty function
+	// ~ 37% of all smeshers won't find a nonce while computing leaves
+	// ~  2% of all smeshers won't find a nonce even after checking 4x numLabels
+	for i := *init.lastPosition; i < math.MaxUint64; i += batchSize {
+		defer init.saveMetadata()
```
moshababo: This will be called (in a reversed order) at the end of the function, not the loop clause. 

How about doing `go init.saveMetadata()` instead? It's best-effort only anyway.

moshababo: It doesn't seem that you are updating `lastPosition` according to the latest PoW-only batch. So upon recovery, you will repeat the same unsuccessful positions.

fasmat: I moved `defer init.saveMetadata()` outside of the for loop. `go init.saveMetadata()` would be racy here -> `init.lastPosition` and `init.nonce` might update between calling `go init.saveMetadata()` and it executing.

The reason I'm using `defer` is because I want to update the metadata file no matter how the loop is ended; by context cancellation, when an error occurs or when the nonce is found.

fasmat: Updated accordingly and added asserts in `ContinueWithPow` tests for the changes.

moshababo: Looks fine now, together with https://github.com/spacemeshos/post/pull/67#discussion_r1034439363.

moshababo (approved):

PR #68, opened 2022-10-12
Use golangci-lint instead of golint

fasmat:
`golint` has been deprecated for a while. In https://github.com/spacemeshos/go-spacemesh we already migrated to `golangci-lint` instead.

This PR updates `post` to also use `golangci-lint` and fixes all complaints that `golangci-lint` has.


.golangci.yml
```diff
@@ -0,0 +1,262 @@
...
+  print-issued-lines: true
+
+  # print linter name in the end of issue text, default is true
+  print-linter-name: true
+
+# all available settings of specific linters
+linters-settings:
+  staticcheck:
+    # Select the Go version to target. The default is '1.13'.
+    go: "1.18"
```
pigmej: Shouldn't we run on 1.19?

countvonzero: go-spacemesh is still using 1.18.
the guiding principal (prior to me joining) is that there should be a good reason (feature we want to use) to upgrade go version. but i don't really feel too strongly.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,7 @@
 name: CI
 
 env:
-  go-version: '1.19'
+  go-version: "1.19"
```
countvonzero, line 4: but why is this 1.19?

fasmat: I tried 1.18. Using it breaks the Windows tests with the following errors:

`ThreadSanitizer failed to allocate 0x000000959000 (9801728) bytes at 0x200d8fbb54000 (error code: 87)`

See e.g. here: https://github.com/spacemeshos/post/actions/runs/3236096321/jobs/5301430586 with 1.19 this does not happen. I don't know if this is a go 1.18 issue or caused by the GCC that is used by Github. See also https://github.com/golang/go/issues/46099

.golangci.yml
fasmat: post breaks on windows when I try to use 1.18. I will update this config to use 1.19 instead.

fasmat: I created a PR to also update `go-spacemesh` to use Go 1.19: https://github.com/spacemeshos/go-spacemesh/pull/3650

pigmej: @countvonzero my point was that in a lot of other places here 1.19 was used ;)

pigmej (approved):
I'm ok with that with the condition that https://github.com/spacemeshos/go-spacemesh/pull/3650 is also ok/merged.

PR #70, opened 2022-10-12
Use gotestsum and staticcheck

fasmat:
This uses gotestsum as test runner and executes staticcheck during every build.

This PR also includes fixes for all complaints by staticcheck.


config/config.go
```diff
@@ -36,7 +36,12 @@ const (
 	MinNumFiles = 1
 )
 
-var DefaultDataDir = filepath.Join(smutil.GetUserHomeDirectory(), "post", DefaultDataDirName)
+var DefaultDataDir string
+
+func init() {
+	home, _ := os.UserHomeDir()
+	DefaultDataDir = filepath.Join(home, "post", DefaultDataDirName)
+}
```
poszu, line 44: What's the benefit of using `init()` here?

config/config.go
```diff
@@ -36,7 +36,12 @@ const (
 	MinNumFiles = 1
 )
 
-var DefaultDataDir = filepath.Join(smutil.GetUserHomeDirectory(), "post", DefaultDataDirName)
+var DefaultDataDir string
+
+func init() {
+	home, _ := os.UserHomeDir()
```
poszu, line 42: Perhaps it could `panic` if it failed?

initialization/initialization_test.go
```diff
@@ -5,50 +5,60 @@ import (
...
 	"github.com/spacemeshos/post/shared"
-	smlog "github.com/spacemeshos/smutil/log"
 	"github.com/stretchr/testify/require"
 )
 
 var (
-	id   = make([]byte, 32)
-	cfg  config.Config
-	opts config.InitOpts
+	id = make([]byte, 32)
```
poszu: `id` could be "ungloballed" too. Consider something like:

```go
func getId() [32]byte {
	return [32]byte{}
}
```

initialization/initialization_test.go
```diff
@@ -5,50 +5,60 @@ import (
...
+	if *logFlag {
+		log.Printf("\tINFO\t"+msg, args...)
+	}
+}
+
+func (l testLogger) Debug(msg string, args ...interface{}) {
+	if *debugFlag {
+		log.Printf("\tDEBUG\t"+msg, args...)
+	}
 }
```
poszu: I feel this should be easier... Why not just use `log.Info()`, `log.Debug()` etc.?

initialization/initialization_test.go
```diff
@@ -63,11 +73,10 @@ func TestInitialize(t *testing.T) {
 func TestInitialize_Repeated(t *testing.T) {
 	r := require.New(t)
 
+	cfg, opts := getTestConfig(t)
 	init, err := NewInitializer(cfg, opts, id)
 	r.NoError(err)
-	if *log {
-		init.SetLogger(smlog.AppLog)
-	}
+	init.SetLogger(testLogger{})
```
poszu: Shouldn't it be part of a `TestMain()` or `init()`?

proving/proving_test.go
```diff
@@ -5,65 +5,73 @@ import (
...
+	if *logFlag {
+		log.Printf("\tINFO\t"+msg, args...)
+	}
+}
+
+func (l testLogger) Debug(msg string, args ...interface{}) {
+	if *debugFlag {
+		log.Printf("\tDEBUG\t"+msg, args...)
+	}
 }
```
poszu: If you stick with this, then I guess it deserves to be factored out to some `test_helpers` pkg.

config/config.go
fasmat: `os.UserHomeDir` returns 2 values, only one of which I'm interested in. Without `init()` I would need to make `home` a global variable too.

config/config.go
fasmat: I intentionally didn't, the code I replaced returned `""` if it was unable to determine the users home directory without panicing. Since this is what `os.UserHomeDir()` does too in the case of an error I just chose to ignore the `error`.

initialization/initialization_test.go
fasmat: I was on the fence of just inlining it. I will create an `id` variable in every test that needs one. I'm opposed to use a function here, since I see no benefit of

```golang
id := getId()
```

over

```golang
id := make([]byte, 32)
```

initialization/initialization_test.go
fasmat: The `log` package doesn't have `log.Info()` and `log.Debug()`: https://pkg.go.dev/log

This logger also isn't used unless explicitly set during the test run by `go test -log ./...` and/or `go test -debug ./...`.

I'm not sure why this is here in the first place; probably to debug failing tests. I don't see a need for this to be honest, I just didn't want to completely remove it (yet).

initialization/initialization_test.go
fasmat: No, `init` is a new instance during every test and must not be re-used for the next test otherwise we are sharing state between tests that shouldn't be shared.

proving/proving_test.go
fasmat: I am strongly against this. With this PR I actually did the exactly the opposite; I got rid of the `test_helpers` pkg (called https://github.com/spacemeshos/smutil) because this is code that should not to be shared.

There is nothing wrong with having your own implementation of the `shared.Logger` interface in every package. It just happens to be the same here as in the other package.

The https://github.com/spacemeshos/smutil package now serves no purpose, afaik no one is using it any more. If you look into it it contains 1:1 the code we JUST removed from the `filesystem` package in go-spacemesh. This is what happens to "util" code that is factored out for re-usage: it is used in exactly one place and forgotten about.

fasmat: The logic is so trivial it doesn't deserve to be it's own package (with documentation of its API and so on!).

If it helps see `testLogger` as a mock of the `shared.Logger` interface. Now pulling in `gomock` is more effort or introduces more complexity than having this small handwritten mock.

initialization/initialization_test.go
poszu: I haven't thought this needs to be that complicated. In other languages one would just bring in a "standard" logger or just print in tests and pass a standard flag like "--no-capture" to test execution.

proving/proving_test.go
poszu: It's not the kind of code you would expect to be different in each pkg. I would actually expect to have consistent logging in tests in various pkgs.

config/config.go
poszu: I'm not sure if the replaced code was right about this. According to the [docs](https://pkg.go.dev/os#UserHomeDir):
> On Unix, including macOS, it returns the $HOME environment variable. On Windows, it returns %USERPROFILE%. On Plan 9, it returns the $home environment variable.

Is just using `""` actually **OK**? It might end up treating the current directory as "home". This will lead to messing up whatever working directory is at the moment. I find it wrong and dangerous. IMO, this is an unexpected situation and should be handled somehow (a panic, a warning log + using /tmp/ etc).

fasmat: Imho raising a panic prevents a usecase that might be important to some:

A smesher decides to run go-spacemesh as daemon with a user that is only used for that process and doesn't have a home directory, password, configured shell, etc. (as it is customary for daemons). The user sets the datadir explicitly for go-spacemesh so it writes to `/opt/spacemesh/` (or something similar). If we add an error check here the program stops not because the data directory is invalid, but because the **default** dataDir couldn't be determined. Logging in this case would also give not valuable information (why is not being able to determine a path that isn't used noteworthy to be added to the logs?)

If the smesher forgets to set a dataDir it would default to `/post/data` (or the equivalent on non-linux systems). If they are running it with root permissions then go-spacemesh just writes there (we don't babysit admins :wink:) if they don't give it root permissions it will fail to create the directory or access files within and tell them so: The first time probably when go-spacemesh tries to load / create the identify file and isn't able to.

If we default to `/tmp` this might cause other issues: go-spacemesh might be running on a system that doesn't have a `/tmp`. `panic` would prevent the user from starting the application at all even if they explicitly set the DataDir.

proving/proving_test.go
fasmat: I would say otherwise. The interface of the logger isn't set by the tests, it's set by the post `Initializer`.

So the package defining that interface would need to provide an implementation for tests. How do you then use the same logger for two different components with different interfaces?

poszu (approved):

PR #75, opened 2022-10-25
Rename ID to Commitment

fasmat:
Since PoST is not using the ID of the node any more as the seed for data creation, rename the parameter used to Commitment instead.

Additionally removed util functions that are not actually used in the library.


initialization/initialization_test.go
```diff
@@ -299,15 +299,15 @@ func TestValidateMetadata(t *testing.T) {
...
+	// Attempt to initialize with different `Commitment`.
 	newID := make([]byte, 32)
 	newID[0] = newID[0] + 1
 	init, err = NewInitializer(cfg, opts, newID)
 	r.NoError(err)
 	err = init.Initialize()
 	errConfigMismatch, ok := err.(ConfigMismatchError)
 	r.True(ok)
-	r.Equal("ID", errConfigMismatch.Param)
+	r.Equal("Commitment", errConfigMismatch.Param)
```
poszu, line 309: `errors.As()` could be used instead of casting ;)

poszu (approved):

PR #76, opened 2022-10-25
Fix possible race in initializer

fasmat:
This fixes an issue in `Initialize()` where `numLabelsWrittenChan` can be set by two goroutines without protection through a mutex:

- `defer` in `Initialize()` is not protected
- `SessionNumLabelsWrittenChan` is possibly called by a different go routine

leading to a possible race. Furthermore the returned channel from `SessionNumLabelsWrittenChan` might never be written to (if called after `Initialize()` returns). Tests occasionally time out for this reason.

Additionally `numLabelsWrittenChan` is a buffered channel to which a blocking send happens in `updateSessionNumLabelsWritten`. This causes the `Initializer` to halt until another goroutine starts reading from the channel. To fix this issue `Initializer` will now skip writing to `numLabelsWrittenChan` if no reader is waiting for updates.

### DIFF
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -91,22 +91,18 @@ func (init *Initializer) Initialize() error {
 
 	init.stopChan = make(chan struct{})
 	init.doneChan = make(chan struct{})
+	init.numLabelsWrittenChan = make(chan uint64)
 
 	init.initializing = true
-
 	init.mtx.Unlock()
 
 	defer func() {
 		init.mtx.Lock()
 		init.initializing = false
-		init.mtx.Unlock()
 
 		close(init.doneChan)
-
-		if init.numLabelsWrittenChan != nil {
-			close(init.numLabelsWrittenChan)
-			init.numLabelsWrittenChan = nil
-		}
+		close(init.numLabelsWrittenChan)
+		init.mtx.Unlock()
 	}()
 
 	if numLabelsWritten, err := init.diskState.NumLabelsWritten(); err != nil {
@@ -143,7 +139,6 @@ func (init *Initializer) Initialize() error {
 func (init *Initializer) isInitializing() bool {
 	init.mtx.RLock()
 	defer init.mtx.RUnlock()
-
 	return init.initializing
 }
 
@@ -167,13 +162,8 @@ func (init *Initializer) Stop() error {
 }
 
 func (init *Initializer) SessionNumLabelsWrittenChan() <-chan uint64 {
-	init.mtx.Lock()
-	defer init.mtx.Unlock()
-
-	if init.numLabelsWrittenChan == nil {
-		init.numLabelsWrittenChan = make(chan uint64, 1024)
-	}
-
+	init.mtx.RLock()
+	defer init.mtx.RUnlock()
 	return init.numLabelsWrittenChan
 }
 
@@ -286,14 +276,13 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
 	ioError := make(chan error, 1)
 
 	var wg sync.WaitGroup
-	wg.Add(2)
 
 	// Start compute worker.
+	wg.Add(1)
 	go func() {
-		defer func() {
-			close(outputChan)
-			wg.Done()
-		}()
+		defer wg.Done()
+		defer close(outputChan)
+
 		for currentPosition < fileNumLabels {
 			select {
 			case <-init.stopChan:
@@ -329,6 +318,7 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
 	}()
 
 	// Start IO worker.
+	wg.Add(1)
 	go func() {
 		defer wg.Done()
 		for {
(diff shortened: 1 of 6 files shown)


initialization/initialization.go
```diff
@@ -286,14 +276,13 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
 	ioError := make(chan error, 1)
 
 	var wg sync.WaitGroup
-	wg.Add(2)
 
 	// Start compute worker.
+	wg.Add(1)
```
poszu: An `errgroup` could be used - it should simplify the code. The "error channels" would become redundant and the `wg` is awaited at the bottom anyway.

initialization/initialization.go
```diff
@@ -362,19 +352,17 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
 	}
 
 	init.logger.Info("initialization: file #%v completed; number of labels written: %v", fileIndex, numLabelsWritten)
-
 	return nil
 }
 
 func (init *Initializer) updateSessionNumLabelsWritten(numLabelsWritten uint64) {
 	atomic.StoreUint64(&init.numLabelsWritten, numLabelsWritten)
```
poszu: I know it's not your code, but as you are already at it... 😁 

If `init.numLabelsWritten` is supposed to be atomic, why the code doesn't make it an atomic type? 🤔

initialization/initialization.go
```diff
@@ -362,19 +352,17 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
...
 
-	init.mtx.RLock()
-	ch := init.numLabelsWrittenChan
-	init.mtx.RUnlock()
-
-	if ch != nil {
-		ch <- numLabelsWritten
+	select {
+	case init.numLabelsWrittenChan <- numLabelsWritten:
```
poszu, line 351: This write will `panic` when `init.numLabelsWrittenChan` is closed. It can happen because closing is asynchronous. IMO, sending should be wrapped with a `init.mtx.Lock()`.

initialization/initialization.go
```diff
@@ -167,13 +162,8 @@ func (init *Initializer) Stop() error {
...
-	init.mtx.Lock()
-	defer init.mtx.Unlock()
-
-	if init.numLabelsWrittenChan == nil {
-		init.numLabelsWrittenChan = make(chan uint64, 1024)
-	}
-
+	init.mtx.RLock()
+	defer init.mtx.RUnlock()
 	return init.numLabelsWrittenChan
```
poszu, line 169: Why is a lock required here?

initialization/initialization.go
```diff
@@ -91,22 +91,18 @@ func (init *Initializer) Initialize() error {
...
-		init.mtx.Unlock()
 
 		close(init.doneChan)
-
-		if init.numLabelsWrittenChan != nil {
-			close(init.numLabelsWrittenChan)
-			init.numLabelsWrittenChan = nil
-		}
+		close(init.numLabelsWrittenChan)
+		init.mtx.Unlock()
```
poszu: NIT, you could use `defer`.

poszu (commented):
`Initializer` definitely requires some tests targeted at thread-safety. Why don't Uou write a test covering the issue that You found?

initialization/initialization.go
fasmat: `defer` in a `defer func() {}`?

image

initialization/initialization.go
fasmat: Creation of the channel is secured by a lock, so accessing it should also be imho.

initialization/initialization.go
fasmat: I'll see if using an `errGroup` makes the code cleaner.

initialization/initialization.go
fasmat: Don't know, but I'll update it

initialization/initialization.go
fasmat: I'll double check if this causes any possible deadlocks, if not I'll add the protection via mutex

fasmat: A call to `Stop()` triggers the closing of this channel. This happens after all go routines in `initFile` have returned, and thereby all possible callers of `updateSessionNumLabelsWritten` have ended.

initialization/initialization.go
poszu: It leads to the conclusion that if `SessionNumLabelsWrittenChan()` is called before the channel is created in `Initialize()`, the caller will receive `nil` or a closed channel (from a previous run). I looked up how this is used in `PostSetupManager` in go-spacemesh`. It is exposed via the `PostSetupProvider` interface and then used from the GRPC API. It will break if the GRPC API user call `PostSetupStatusStream()` before calling `StartSmeshing`.

IMO, `Initializer` (and `PostSetupManager`) could be made much simpler and robust if it was not made "reusable". 

`PostSetupManager::StartSession()` already creates a completely new instance of `Initializer` rather than reusing the old one. 

`Initializer::Stop` could be changed to just cancelling a context.

I'm not sure what `Initializer::Reset()` is for.

fasmat: It looks like `Reset()` is used primarily in tests to clean up between tests and by the PostSetupManager to delete the local PoST files if necessary. This is currently only the case in tests as well.

The only use case I see for calling `Reset()` on a live node, is if the node changes identities. In this case it might be better to just have multiple folders, one for each identity, so that changing identities doesn't require re-initialising the PoST for them.

I agree with the point of `Stop()` being unnecessary if we would just pass in a `context` to `Initialize()` and cancel that if we want to stop the Initializer from what it is doing.

Regarding `SessionNumLabelsWrittenChan()`: imho this function isn't needed at all. `PostSetupManager::StatusChan()` should just call `SessionNumLabelsWritten()` and `Started()` `Completed()`, `isInitializing()` or better a single method that combines the purpose of all 3. `PostSetupmanager::StatusChan()` could then just return the current status instead of a channel, which in turn is called by `SmesherService::PostSetupStatusStream()` in regular intervals.

This way `PostSetupStatusStream()` controls how often status updates are sent via the stream instead of `Initializer`. All this methods should return quickly and not interfere with initialization via blocking of channels or similar.

I'll create a ticket for the refactoring, since this should be done separately from this PR.

poszu: LGTM

poszu (approved):

Issue #78, opened 2022-10-28
Refactor status updates by PoST Initializer


fasmat:
Based on the discussion in these two comments:
- https://github.com/spacemeshos/post/pull/76#discussion_r1007295465
- https://github.com/spacemeshos/post/pull/76#discussion_r1007320896

At the moment the `Initializer` type is overly complex in its usage. It is designed to be re-used, when it doesn't need to be and over-uses channels and go routines. The type can be simplified in the following manner:

- Simplify `Initializer` by making it non-reusable; besides tests the type isn't used in this manner anyway
- Remove `Reset()`: it is only used in tests and the directory used by `Initializer` is known to the caller. In tests the caller can delete the directory itself. In Production this shouldn't be done anyway; switching identities should instead use a different datadir for PoST.
- Remove `SessionNumLabelsWrittenChan()`; there is no need for this method when `SessionNumLabelsWritten()` returns the same information in a synchronous way. Callers interested in updates should rely on the values returned by the second method and the status from `Started()`, `Completed()`, `isInitializing()` or even a new single method that combines the purpose of all 3.
- `Initialize()` should take a `context` as parameter and instead of `Stop()` callers should just cancel the context they passed in.

Refactorings of types using `Initializer` in go-spacemesh:

- `PostSetupManager`:
   - `StatusChan()` should just call `SessionNumLabelsWritten()` and return the status directly instead of via a channel
-  `SmesherService`:
   - `PostStetupStatusStream()` should instead of relying on receiving values in regular intervals via the channel, call the (new) synchronous method in `PostSetupManager` in regular intervals.

fasmat:
@poszu would you add/change something on this ticket?

PR #83, opened 2022-11-06
Fix incorrect count of labels written

fasmat:
This is a fix for a bug that was introduced in this commit: https://github.com/spacemeshos/post/commit/55a0a5115ea2f606816a05523fde4b90075964df#diff-6682f153c20b292747e05c02ab1907c2d29c9888a086349448302f85f31ee6b9L348-R349

Instead of updating the number of labels written, the new value is incorrectly added to the current value, causing `Initializer` to report an incorrect number of the total.

### DIFF
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -345,7 +345,7 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
 }
 
 func (init *Initializer) updateSessionNumLabelsWritten(numLabelsWritten uint64) {
-	init.numLabelsWritten.Add(int64(numLabelsWritten))
+	init.numLabelsWritten.Store(int64(numLabelsWritten))
 
 	select {
 	case init.numLabelsWrittenChan <- numLabelsWritten:
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -46,7 +46,7 @@ func CPUProviderID() int {
 }
 
 type Initializer struct {
-	numLabelsWritten     atomic.Int64
+	numLabelsWritten     atomic.Uint64
 	numLabelsWrittenChan chan uint64
 
 	cfg        Config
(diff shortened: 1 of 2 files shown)


initialization/initialization.go
```diff
@@ -345,7 +345,7 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
 }
 
 func (init *Initializer) updateSessionNumLabelsWritten(numLabelsWritten uint64) {
-	init.numLabelsWritten.Add(int64(numLabelsWritten))
+	init.numLabelsWritten.Store(int64(numLabelsWritten))
```
poszu: `Uint64` was added in go 1.19: https://pkg.go.dev/sync/atomic#Uint64.

I suggest changing the type of `init.numLabelsWritten` to `atomic.Uin64` and simplifying accesses to:
```suggestion
	init.numLabelsWritten.Store(numLabelsWritten)
```

initialization/initialization_test.go
```diff
@@ -49,9 +49,9 @@ func TestInitialize(t *testing.T) {
 	r.NoError(err)
 	init.SetLogger(log)
 
-	eg, _ := errgroup.WithContext(context.Background())
+	var eg errgroup.Group
+	eg.Go(init.Initialize)
```
poszu: Why does it need to run asynchronously?

initialization/initialization_test.go
```diff
@@ -443,6 +443,10 @@ func Test_SessionNumLabelsWrittenChan_Racefree(t *testing.T) {
 
 func assertNumLabelsWrittenChan(init *Initializer, t *testing.T) func() error {
 	return func() error {
+		// hack to avoid receiving a nil channel from SessionNumLabelsWrittenChan()
+		// TODO (mafa): for a proper fix see https://github.com/spacemeshos/post/issues/78
+		time.Sleep(10 * time.Millisecond)
+
```
poszu: Kudos for the TODO with a link ❤️.

Perhaps you could use `require.Eventually(...)` with a longer timeout for receiving the channel instead? It would be less flaky than this.

poszu (changes requested):
A request to change to an `atomic.Uint64`.

initialization/initialization_test.go
fasmat: This is already in preparation of `init.Initialize()` becoming a blocking call in the future as outlined in https://github.com/spacemeshos/post/issues/78

poszu: Then, IMO, it should be done with the changes for #78.

poszu (approved):

PR #84, opened 2022-11-07
Refactor initializer to be simpler to use and race free

### LINKED ISSUE #78: Refactor status updates by PoST Initializer

fasmat:
This fixes #78 by slimming down the `Initializer` and removing asynchronous calls. This should also fix data race issues we have with the component.

I did not remove the `Reset` method, since it is used by `go-spacemesh` and factoring it out would have been more effort than what I think is worth at the moment.

Now `Initializer::Initialize()` takes a context that can be canceled by the caller to stop the initialization early. `Initializer::Stop()` was removed as it is not needed any more and instead of having multiple methods that report some kind of state (`Started()`, `Completed()`, etc.) there is now only one method called `Status()` that reports the state of the initialization.

For integration into go-spacemesh see: https://github.com/spacemeshos/go-spacemesh/pull/3715

### DIFF
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -46,8 +46,7 @@ func CPUProviderID() int {
 }
 
 type Initializer struct {
-	numLabelsWritten     atomic.Uint64
-	numLabelsWrittenChan chan uint64
+	numLabelsWritten atomic.Uint64
 
 	cfg        Config
 	opts       InitOpts
@@ -93,7 +92,6 @@ func (init *Initializer) Initialize() error {
 
 	init.stopChan = make(chan struct{})
 	init.doneChan = make(chan struct{})
-	init.numLabelsWrittenChan = make(chan uint64)
 
 	init.initializing = true
 	init.mtx.Unlock()
@@ -104,7 +102,6 @@ func (init *Initializer) Initialize() error {
 		init.initializing = false
 
 		close(init.doneChan)
-		close(init.numLabelsWrittenChan)
 	}()
 
 	if numLabelsWritten, err := init.diskState.NumLabelsWritten(); err != nil {
@@ -248,7 +245,7 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
 	if numLabelsWritten > 0 {
 		if numLabelsWritten == fileNumLabels {
 			init.logger.Info("initialization: file #%v already initialized; number of labels: %v, start position: %v", fileIndex, numLabelsWritten, fileOffset)
-			init.updateSessionNumLabelsWritten(fileTargetPosition)
+			init.numLabelsWritten.Store(fileTargetPosition)
 			return nil
 		}
 
@@ -257,7 +254,7 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
 			if err := writer.Truncate(fileNumLabels); err != nil {
 				return err
 			}
-			init.updateSessionNumLabelsWritten(fileTargetPosition)
+			init.numLabelsWritten.Store(fileTargetPosition)
 			return nil
 		}
 
@@ -301,7 +298,7 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
 			outputChan <- output
 			currentPosition += uint64(batchSize)
 
-			init.updateSessionNumLabelsWritten(fileOffset + currentPosition)
+			init.numLabelsWritten.Store(fileOffset + currentPosition)
 		}
 		return nil
 	})
@@ -336,17 +333,6 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
 
 	init.logger.Info("initialization: file #%v completed; number of labels written: %v", fileIndex, numLabelsWritten)
 	return nil
-}
-
-func (init *Initializer) updateSessionNumLabelsWritten(numLabelsWritten uint64) {
-	init.numLabelsWritten.Store(numLabelsWritten)
-
-	select {
-	case init.numLabelsWrittenChan <- numLabelsWritten:
-	default:
-		// if no one listens for the update, we just drop it
-		// otherwise Initializer would eventually stop working until someone reads from the channel
-	}
 }
 
 func (init *Initializer) verifyMetadata(m *Metadata) error {
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -163,12 +163,6 @@ func (init *Initializer) Stop() error {
 	return nil
 }
 
-func (init *Initializer) SessionNumLabelsWrittenChan() <-chan uint64 {
-	init.mtx.RLock()
-	defer init.mtx.RUnlock()
-	return init.numLabelsWrittenChan
-}
-
 func (init *Initializer) SessionNumLabelsWritten() uint64 {
 	return init.numLabelsWritten.Load()
 }
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -9,7 +9,6 @@ import (
 	"path/filepath"
 	"sync"
 	"sync/atomic"
-	"time"
 
 	"golang.org/x/sync/errgroup"
 
@@ -29,11 +28,19 @@ type (
 	ComputeProvider     = gpu.ComputeProvider
 )
 
+type Status int
+
+const (
+	StatusNotStarted Status = iota
+	StatusStarted
+	StatusInitializing
+	StatusCompleted
+	StatusError
+)
+
 var (
-	ErrNotInitializing              = errors.New("not initializing")
 	ErrAlreadyInitializing          = errors.New("already initializing")
 	ErrCannotResetWhileInitializing = errors.New("cannot reset while initializing")
-	ErrStopped                      = errors.New("gpu-post: stopped")
 	ErrStateMetadataFileMissing     = errors.New("metadata file is missing")
 )
 
@@ -43,6 +50,86 @@ func Providers() []ComputeProvider {
 
 func CPUProviderID() int {
 	return gpu.CPUProviderID()
+}
+
+func RemoveDataFiles(dataDir string) error {
+	files, err := os.ReadDir(dataDir)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		info, err := file.Info()
+		if err != nil {
+			continue
+		}
+		if shared.IsInitFile(info) || file.Name() == metadataFileName {
+			path := filepath.Join(dataDir, file.Name())
+			if err := os.Remove(path); err != nil {
+				return fmt.Errorf("failed to delete file (%v): %w", path, err)
+			}
+		}
+	}
+
+	return nil
+}
+
+type initializeOption struct {
+	commitment []byte
+	cfg        *Config
+	initOpts   *config.InitOpts
+	logger     Logger
+}
+
+func (opts *initializeOption) verify() error {
+	if opts.cfg == nil {
+		return errors.New("no config provided")
+	}
+
+	if opts.initOpts == nil {
+		return errors.New("no init options provided")
+	}
+
+	if err := config.Validate(*opts.cfg, *opts.initOpts); err != nil {
+		return err
+	}
+	return nil
+}
+
+type initializeOptionFunc func(*initializeOption) error
+
+func WithCommitment(commitment []byte) initializeOptionFunc {
+	return func(opts *initializeOption) error {
+		if len(commitment) != 32 {
+			return fmt.Errorf("invalid `id` length; expected: 32, given: %v", len(commitment))
+		}
+		opts.commitment = commitment
+		return nil
+	}
+}
+
+func WithInitOpts(initOpts config.InitOpts) initializeOptionFunc {
+	return func(opts *initializeOption) error {
+		opts.initOpts = &initOpts
+		return nil
+	}
+}
+
+func WithConfig(cfg Config) initializeOptionFunc {
+	return func(opts *initializeOption) error {
+		opts.cfg = &cfg
+		return nil
+	}
+}
+
+func WithLogger(logger Logger) initializeOptionFunc {
+	return func(opts *initializeOption) error {
+		if logger == nil {
+			return errors.New("logger is nil")
+		}
+		opts.logger = logger
+		return nil
+	}
 }
 
 type Initializer struct {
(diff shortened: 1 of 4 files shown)


initialization/initialization.go
```diff
@@ -45,9 +52,66 @@ func CPUProviderID() int {
...
+	}
+
+	if opts.initOpts == nil {
+		return errors.New("no init options provided")
+	}
+
+	if err := config.Validate(*opts.cfg, *opts.initOpts); err != nil {
+		return err
+	}
+	return nil
```
poszu:
```suggestion
	return config.Validate(*opts.cfg, *opts.initOpts)
```

initialization/initialization.go
```diff
@@ -275,17 +300,22 @@ func (init *Initializer) initFile(computeProviderID uint, fileIndex int, numLabe
...
 		defer close(outputChan)
 
 		for currentPosition < fileNumLabels {
 			select {
-			case <-init.stopChan:
+			case <-ctx.Done():
 				init.logger.Info("initialization: stopped")
-				return ErrStopped
+
+				if res := gpu.Stop(); res != gpu.StopResultOk {
```
poszu, line 292: Is it OK to call `gpu.Stop()` only when the context is cancelled? Doesn't it need to be called after finishing normally too?

initialization/initialization.go
```diff
@@ -144,42 +204,16 @@ func (init *Initializer) isInitializing() bool {
...
-	if init.isInitializing() {
+	switch init.Status() {
+	case StatusInitializing:
 		return ErrCannotResetWhileInitializing
-	}
-
-	if err := init.VerifyStarted(); err != nil {
-		return err
+	case StatusError:
+		return fmt.Errorf("cannot determine status of initialization")
```
poszu: `Reset()` and `Status()` are meant to be called concurrently to `Initialize()`, right?

The status might change right after this check. Consider the following scenario. A and B are different goroutines.
1. A calls `init.Reset()`, `init.Status() != StatusInitializing` 
2. B calls `init.Initialize()`. it starts running
3. A continues to execute `Reset()`. Now the requirement that initializer is not running during `Reset()` is violated.

initialization/initialization.go
```diff
@@ -203,38 +237,29 @@ func (init *Initializer) Reset() error {
...
 	}
-	if !ok {
-		return shared.ErrInitNotStarted
+
+	if numLabelsWritten > 0 {
+		return StatusStarted
 	}
 
-	return nil
+	return StatusNotStarted
```
poszu, line 237: Given these conditions `StatusStarted` means "already produced something" and `StatusNodStarted` means "not produced anything yet (but might be started already)".

initialization/initialization_test.go
```diff
@@ -41,425 +46,466 @@ func (l testLogger) Debug(msg string, args ...interface{}) { l.t.Logf("\tDEBUG\t
...
 
-	// Cleanup.
-	r.NoError(init.Reset())
+		var eg errgroup.Group
+		eg.Go(assertNumLabelsWritten(ctx, t, init))
+		r.NoError(init.Initialize(ctx))
+		cancel()
+		eg.Wait()
+	}
 }
```
poszu: I don't understand this test. How is it verifying that Initializer actually does something? The checks in `assertNumLabelsWritten()` only check if `init.SessionNumLabelsWritten()` returns values that are monotonically nondecreasing. The initializer could just hang and do nothing and return 0 all the time. Am I missing something?

initialization/initialization_test.go
```diff
@@ -41,425 +46,466 @@ func (l testLogger) Debug(msg string, args ...interface{}) { l.t.Logf("\tDEBUG\t
...
 
 	// Initialize.
-	err = init.Initialize()
-	req.NoError(err)
+	r.NoError(init.Initialize(context.Background()))
 	numLabelsWritten, err = init.diskState.NumLabelsWritten()
-	req.NoError(err)
-	req.Equal(uint64(opts.NumUnits)*cfg.LabelsPerUnit, numLabelsWritten)
+	r.NoError(err)
+	r.Equal(uint64(init.opts.NumUnits)*init.cfg.LabelsPerUnit, numLabelsWritten)
```
poszu: IMO, the test should not access the private fields of the tested thing. Why not check against local `opts`?

initialization/initialization_test.go
```diff
@@ -41,425 +46,466 @@ func (l testLogger) Debug(msg string, args ...interface{}) { l.t.Logf("\tDEBUG\t
...
+						cancel()
+						return nil
+					}
+				}
 			}
-		}
-		return nil
-	})
-	r.NoError(eg.Wait())
+		})
```
poszu: This whole block inside `eg.Go()` could be rewritten to just `r.Eventually()` with a reasonably big timeout, like:
```go
r.Eventually(func() bool { return init.SessionNumLabelsWritten() > 0}, time.Minute, time.Millisecond * 50)
cancel()
```

initialization/initialization_test.go
```diff
@@ -41,425 +46,466 @@ func (l testLogger) Debug(msg string, args ...interface{}) { l.t.Logf("\tDEBUG\t
...
-	// Cleanup.
-	r.NoError(init.Reset())
+	// Continue the initialization to completion.
+	{
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		var eg errgroup.Group
+		eg.Go(assertNumLabelsWritten(ctx, t, init))
+		r.NoError(init.Initialize(ctx))
```
poszu, line 546: During our call, I understood that `Initializer` cannot be used twice. Can it?

initialization/initialization_test.go
```diff
@@ -41,425 +46,466 @@ func (l testLogger) Debug(msg string, args ...interface{}) { l.t.Logf("\tDEBUG\t
...
-			t.Logf("num labels written: %v\n", p)
+
+		for {
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-timer.C:
+				num := init.SessionNumLabelsWritten()
+				t.Logf("num labels written: %v\n", num)
+				assert.LessOrEqual(t, prev, num)
```
poszu: The intent is to check if `num` is increasing over time. IMO, this check should be inverted to match the intent :) something like:
```suggestion
				assert.GreaterOrEqual(t, num, prev)
```

proving/proving_test.go
```diff
@@ -51,54 +53,60 @@ func TestProver_GenerateProof(t *testing.T) {
...
-		log.Info("numLabels: %v, indices size: %v\n", numLabels, len(proof.Indices))
-
-		err = verifying.Verify(proof, proofMetaData)
-		r.NoError(err)
-
-		// Cleanup.
-		err = init.Reset()
-		r.NoError(err)
+		t.Run(fmt.Sprintf("numUnits=%d", numUnits), func(t *testing.T) {
+			opts.NumUnits = numUnits
```
poszu: It should be fine to run these iterations in parallel, should it not? 🙃 
```suggestion
		t.Run(fmt.Sprintf("numUnits=%d", numUnits), func(t *testing.T) {
			t.Parallel()
			opts.NumUnits = numUnits
```

In general, probably all the tests should be OK to be run in parallel.

poszu (changes requested):
Great work! I left few comments, some of them are just nits.

initialization/initialization.go
fasmat: I did some digging, `gpu.Stop()` calls this C function with a `ms_timeout` value of 2000: https://github.com/spacemeshos/gpu-post/blob/develop/src/api_internal.cpp#L108

It essentially sets `g_spacemesh_api_abort_flag` to `true` which is used as a signal to terminate ongoing work like e.g. here: https://github.com/spacemeshos/gpu-post/blob/develop/src/cuda/driver-cuda.cpp#L220 (and in a few other places)

If initialization is not interrupted but continues to completion `gpu.Stop()` does not need to be called. A better name for that function would probably be `Abort()` or `Cancel()`

PS.: looking through this code I believe the whole thing is inherently racy. `g_spacemesh_api_abort_flag` is a global variable that is written to / read from different go routines in our application, unless the `volatile` keyword makes that OK I think there is a mutex missing that protects access to `g_spacemesh_api_abort_flag`.

initialization/initialization.go
fasmat: `Reset()` as you say can be a problem... I just remembered the `TryLock()` method of a `sync.Mutex`; I think by locking both `Initialize()` and `Reset()` behind the same mutex and using `TryLock()` over `Lock()` to be able to return early if the mutex is in use by the other method we can solve this concurrency issue.

Maybe this even lets us get rid of the `initializing` field. What's your opinion?

initialization/initialization.go
fasmat: `StatusNotStarted` only means "not produced anything yet (but might be started already)" if `Initialize()` is called between the check of `initialize` at the top of this method and this return.

Here again I think the removal of `initialize` and using `TryLock()` might resolve the issues with the `Status()` method.

initialization/initialization_test.go
fasmat: The call to `init.Initialize()` is synchronous. If it returns without an error, the Initializer has completed its work and calling `init.Status()` would return `StatusCompleted` and `init.SessionNumLabelsWritten()` would return `opts.NumUnits * cfg.LabelsPerUnit`. The later is the condition that is checked for by Initializer to stop its work and the former returns `StatusCompleted` if that condition is met. Both of these conditions are tested for in other tests and are not the focus of this test.

If fewer labels would have been written but Initializer returns early it would return with an error.

initialization/initialization_test.go
fasmat: Depends; in a unit test testing only the Initializer I think it is OK to access private fields. I'm not testing behaviour here, I'm just accessing data. It's not unusual to do this in go (example test from the standard library for `context` where a test checks that a parent context actually contains their children: https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/context/context_test.go;l=158-167)

Testing this only via public methods would either require getter for these values (which we don't want to add just because of a test) or make the test more complicated.

The specific reason I'm doing it here is so that I don't need to re-define `opts` and `cfg` for this test. Both of those are hidden behind `opts := getTestOptions(t)` which returns a slice of functions that I would need to execute to get access to those values. Alternatively I could inline `getTestOptions` (it is only used here and in one other test). Would that be OK for you?

initialization/initialization_test.go
fasmat: That was my original intention; but with the current implementation I see no drawback to having `Initialize` being able to be called twice. The second call returns (almost) immediately if initialisation previously already completed or continues where the previous call left off.

Initially I had the plan to make `Initializer::Initialize()` just a function that receives all the arguments of `NewInitializer` and a `context` that can be cancelled to abort PoST initialization. But this would require a bigger refactoring:

- `Status()` could be completely removed: the caller knows the status, either they didn't call `Initialize()` yet, they called it and it and it hasn't returned yet, they called it and it returned with an error or it completed successfully.
- `Reset()` could be its own function that just receives the `DataDir` as parameter.
- `NumLabelsWritten()` is a bit more involved but could also be a function that only requires `DataDir` as a parameter.

initialization/initialization_test.go
poszu: There are two issues with unit tests accessing private fields.

1. They assume certain implementation. It impedes refactoring.
2. The test uses private data of `init` as the expected value in the assertion. It assumes that `init.opts` is properly assigned in `NewInitializer(opts...)`. If it is not properly assigned because of a bug, then the assertion would be invalid.

I think that inlining `getTestOptions` as you suggested is the way to go 😉 .

initialization/initialization.go
poszu: 😮 The Go wrapper could provide a safe abstraction over the C library. If the C library `gpu-post` has a global state and thus can only be safely called once at the same time (no concurrency), then, IMHO, the Go wrapper around it could prevent calling it concurrently. It could, for example, have a global mutex locked on "start" and unlocked on "stop". Or at least panic if "start" is called twice (IMO panic is better then UB).

Otherwise, it should be at least documented in `gpu/api.go`.

initialization/initialization.go
poszu: I think there are too many things to describe the same property (there are status and initializing), they can go out of sync. A quick fix is to lock `init.mtx` before checking the status to avoid racing with `init.Initialize()` and unlock it upon quitting.

Otherwise, the whole `Initializer` could be documented as not thread-safe and the user should use a Mutex if the user wants to use it concurrently. 

I'd go with the second approach but it requires more refactoring on the go-spacemesh side.

initialization/initialization_test.go
poszu: I still don't understand the merit of this test. It's called `TestInitialize()` yet it passes if I replace the body of the `Initialize()` method with `return nil`.

initialization/initialization.go
fasmat: I created an issue to fix the racy `Stop()` call in a future PR here: https://github.com/spacemeshos/gpu-post/issues/79

initialization/initialization_test.go
fasmat: I added an additional assert that checks if the expected number of labels have been created.

initialization/initialization.go
fasmat: The problem with locking `Status()` and `Initialize()` behind the same mutex is, that `Status()` must be able to be called while `Initialize()` is running.

I know this shouldn't be necessary because the caller of `Status()` should know if `Initialize()` is currently running or not, but changing that behaviour is a bigger effort and out of scope of this PR.

I will slightly improve the current condition by using `TryLock()` instead of `Lock()` to catch if `Status()` should return `StatusInitializing`.

poszu: `Reset()` still has the same issue. Nothing prevents Initialization to start after `init.Status()` returns.

fasmat: Added `TryLock()` to `Reset()` and implemented a test checking the functionality.

poszu (approved):

PR #91, opened 2022-12-06
Make PowDifficulty easier

fasmat:
The current difficulty function for finding a nonce to be used for POPS-VRF is not a good fit. With it more than 1/3 of smeshers are expected to continue initialization when their PoST has been filled.

This reduces the difficulty of finding a nonce such that 99.97% of all smeshers are expected to find at least one index that satisfies the difficulty during initialization. Only 0.01% of smeshers are expected to do check more than 115% of the labels required for PoST initialization with this difficulty function.

### DIFF
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -41,8 +41,6 @@ var (
 	ErrAlreadyInitializing          = errors.New("already initializing")
 	ErrCannotResetWhileInitializing = errors.New("cannot reset while initializing")
 	ErrStateMetadataFileMissing     = errors.New("metadata file is missing")
-
-	powDifficultyFunc = shared.PowDifficulty
 )
 
 // Providers returns a list of available compute providers.
@@ -63,7 +61,9 @@ type option struct {
 
 	cfg      *Config
 	initOpts *config.InitOpts
-	logger   Logger
+
+	logger            Logger
+	powDifficultyFunc func(uint64) []byte
 }
 
 func (o *option) validate() error {
@@ -141,6 +141,18 @@ func WithLogger(logger Logger) OptionFunc {
 	}
 }
 
+// withDifficultyFunc sets the difficulty function for the initializer.
+// NOTE: This is an internal option for tests and should not be used by external packages.
+func withDifficultyFunc(powDifficultyFunc func(uint64) []byte) OptionFunc {
+	return func(opts *option) error {
+		if powDifficultyFunc == nil {
+			return errors.New("difficulty function is nil")
+		}
+		opts.powDifficultyFunc = powDifficultyFunc
+		return nil
+	}
+}
+
 // Initializer is responsible for initializing a new PoST commitment.
 type Initializer struct {
 	nodeId          []byte
@@ -158,12 +170,15 @@ type Initializer struct {
 	diskState        *DiskState
 	mtx              sync.RWMutex
 
-	logger Logger
+	logger            Logger
+	powDifficultyFunc func(uint64) []byte
 }
 
 func NewInitializer(opts ...OptionFunc) (*Initializer, error) {
 	options := &option{
 		logger: shared.DisabledLogger{},
+
+		powDifficultyFunc: shared.PowDifficulty,
 	}
 
 	for _, opt := range opts {
@@ -177,13 +192,14 @@ func NewInitializer(opts ...OptionFunc) (*Initializer, error) {
 	}
 
 	return &Initializer{
-		cfg:             *options.cfg,
-		opts:            *options.initOpts,
-		nodeId:          options.nodeId,
-		commitmentAtxId: options.commitmentAtxId,
-		commitment:      options.commitment,
-		diskState:       NewDiskState(options.initOpts.DataDir, uint(options.cfg.BitsPerLabel)),
-		logger:          options.logger,
+		cfg:               *options.cfg,
+		opts:              *options.initOpts,
+		nodeId:            options.nodeId,
+		commitmentAtxId:   options.commitmentAtxId,
+		commitment:        options.commitment,
+		diskState:         NewDiskState(options.initOpts.DataDir, uint(options.cfg.BitsPerLabel)),
+		logger:            options.logger,
+		powDifficultyFunc: options.powDifficultyFunc,
 	}, nil
 }
 
@@ -215,7 +231,7 @@ func (init *Initializer) Initialize(ctx context.Context) error {
 
 	numLabels := uint64(init.opts.NumUnits) * uint64(init.cfg.LabelsPerUnit)
 	fileNumLabels := numLabels / uint64(init.opts.NumFiles)
-	difficulty := powDifficultyFunc(numLabels)
+	difficulty := init.powDifficultyFunc(numLabels)
 	batchSize := uint64(config.DefaultComputeBatchSize)
 
 	init.logger.Info("initialization: starting to write %v file(s); number of units: %v, number of labels per unit: %v, number of bits per label: %v, datadir: %v",
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -41,6 +41,8 @@ var (
 	ErrAlreadyInitializing          = errors.New("already initializing")
 	ErrCannotResetWhileInitializing = errors.New("cannot reset while initializing")
 	ErrStateMetadataFileMissing     = errors.New("metadata file is missing")
+
+	powDifficultyFunc = shared.PowDifficulty
 )
 
 // Providers returns a list of available compute providers.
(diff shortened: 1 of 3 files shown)


initialization/initialization.go
```diff
@@ -41,6 +41,8 @@ var (
 	ErrAlreadyInitializing          = errors.New("already initializing")
 	ErrCannotResetWhileInitializing = errors.New("cannot reset while initializing")
 	ErrStateMetadataFileMissing     = errors.New("metadata file is missing")
+
+	powDifficultyFunc = shared.PowDifficulty
```
poszu: Please, let's not use mutable globals. It's one of the sources of data races and prevents running tests in parallel.

How about adding a `powDifficulty func (numLabels uint64) []byte` field to `Initializer struct`, setting it to `shared.PowDifficulty` by default and allowing overriding with `OptionFunc`?

shared/shared.go
```diff
@@ -37,18 +37,20 @@ func ProvingDifficulty(numLabels uint64, k1 uint64) uint64 {
...
-// in the first `numLabels`	labels and only a ~ 90% chance to find a nonce in the first `2*numLabels` labels.
+// The probability of finding a label below the difficulty threshold within numLabels
+// approaches ~ 99.97% the bigger numLabels gets. Within 1.15 * numLabels the probability
+// approaches 99.99% of finding at least one label below the difficulty threshold.
 func PowDifficulty(numLabels uint64) []byte {
 	difficulty := make([]byte, 33)
 	difficulty[0] = 0x01
 	x := new(big.Int).SetBytes(difficulty)
 	x.Div(x, big.NewInt(int64(numLabels)))
+	x.Lsh(x, 3) // x << 3 == x * 2^3 == x * 8
```
poszu: The comment wouldn't be necessary if you just multiplied by 8:
```suggestion
	x.Mul(x, 8)
```

poszu (changes requested):
Requesting changes for the mutable global variable.

initialization/initialization.go
fasmat: It should not be possible to modify difficulty outside of tests.

The only reason it exists is to be able to test correct behaviour that should only occur for 0.01% of all users.

EDIT: these tests won't work in parallel anyway, since the Initializer itself doesn't work if accessed by more than one thread (due to the `gpu-post` library, not because of the Go code).

shared/shared.go
fasmat: The reason is speed and memory efficiency:

```
goos: linux
goarch: arm64
pkg: github.com/spacemeshos/post/shared
Benchmark_Lsh-6   	26290420	        45.26 ns/op	     112 B/op	       2 allocs/op
Benchmark_Mul-6   	18627200	        62.79 ns/op	     120 B/op	       3 allocs/op
PASS
ok  	github.com/spacemeshos/post/shared	2.482s
```

poszu: It's sad that the go stdlib doesn't figure out such optimizations itself. We are back in the '90s (manual clever optimizations that compilers do now).

Btw, is it even a hot path? In other words, does this optimization make a real difference? If not - I'd suggest sticking with the obvious `Mul`.

initialization/initialization.go
poszu: The data race in `gpu-post` (which must be fixed anyway) is not an excuse for introducing another.

If you want to restrict the override to be private and used only inside the package (tests), then you can just make the respective field in the options structure private. What do you think?

shared/shared.go
```diff
@@ -37,18 +37,20 @@ func ProvingDifficulty(numLabels uint64, k1 uint64) uint64 {
...
 //
-// TODO(mafa): this difficulty calculation is unfit. There is only a ~ 63% chance to find a nonce
-// in the first `numLabels`	labels and only a ~ 90% chance to find a nonce in the first `2*numLabels` labels.
+// The probability of finding a label below the difficulty threshold within numLabels
+// approaches ~ 99.97% the bigger numLabels gets. Within 1.15 * numLabels the probability
+// approaches 99.99% of finding at least one label below the difficulty threshold.
 func PowDifficulty(numLabels uint64) []byte {
 	difficulty := make([]byte, 33)
 	difficulty[0] = 0x01
 	x := new(big.Int).SetBytes(difficulty)
```
moshababo: Can you explain the last 3 lines? I'm not sure I understand the reason for assigning `0x01` and allocating 33 bytes.

shared/shared.go
moshababo: Can you explain the rationale for reducing difficulty by 8? Did it just happen to fit the desired probability?

shared/shared.go
fasmat: `2^256` (33 bytes) is basically the same as `2^256 - 1` (32 bytes). The former is easier to allocate as number:

```golang
difficulty := make([]byte, 33)
difficulty[0] = 0x01
x := new(big.Int).SetBytes(difficulty)
```

than the later

```golang
difficulty := make([]byte, 32)
for i := range difficulty {
    difficulty[i] = 0xff
}
```

now that I think about it; `x := big.NewInt(1).Lsh(256)` would have the same effect as the 33 bytes with leading 1.

shared/shared.go
fasmat: This was discussed in the last research sync. If we don't reduce the difficulty by a factor of 8 only ~ 2/3 of smeshers will find a nonce during initialization and ~ 3% of smeshers would need to calculate more than 3x `numLabels` to find a nonce that fulfils the difficulty.

When we reduce the difficulty by a factor of 8x ~99.97% of smeshers will find a nonce during initialization and less than 0.01% will need to check more than 1.15x `numLabels`.

This gives an adversary an 8x advantage over an honest miner, but still discourages grinding keys. There have been discussions on how to improve the scheme in the future (see https://github.com/spacemeshos/pm/issues/200), but with POPS-VRF we have to lower the difficulty or honest smeshers will not find a nonce with reasonable effort.

initialization/initialization.go
fasmat: Even if we fix the data race in `gpu-post` I don't think the library can be used to calculate `PoST` data of two different identities or two batches concurrently. So Initializer should be treated as a singleton anyway.

Nevertheless I removed the global variable and made it a private field.

shared/shared.go
moshababo: Thanks!

shared/shared.go
moshababo: Got it, thanks!

moshababo (approved):

initialization/initialization.go
```diff
@@ -141,6 +141,18 @@ func WithLogger(logger Logger) OptionFunc {
...
+// NOTE: This is an internal option for tests and should not be used by external packages.
+func withDifficultyFunc(powDifficultyFunc func(uint64) []byte) OptionFunc {
+	return func(opts *option) error {
+		if powDifficultyFunc == nil {
+			return errors.New("difficulty function is nil")
+		}
+		opts.powDifficultyFunc = powDifficultyFunc
+		return nil
+	}
+}
```
poszu, line 154: I'd just check against nil and panic as this is clearly a logic bug:

```suggestion
func withDifficultyFunc(powDifficultyFunc func(uint64) []byte) OptionFunc {
	if powDifficultyFunc == nil {
		panic("difficulty function is nil")
	}
	return func(opts *option) error {
		opts.powDifficultyFunc = powDifficultyFunc
		return nil
	}
}
```

poszu (approved):

initialization/initialization.go
fasmat: why panic instead of just erroring and falling back to the default difficulty function?

shared/shared.go
fasmat: Yes the stdlib seems to not optimize this. I still feel like anyone working on the codebase should know what a bitshift is and that `x <<= 3` is the same as `x *= 8`

initialization/initialization.go
fasmat: A user might incorrectly assume that `withDifficultyFunc(nil)` they can disable looking for a nonce, analogous as to how it works with `oracle.WorkOracle`'s `WithDifficulty(d []byte)` function.

In this case `NewInitializer` should just return an error instead of panicing imo.

poszu: It's a matter of deciding what is the valid usage of the API. If passing `nil` is clearly wrong, then a `panic` would be easier to debug / faster to catch.

Anyway, IMO, the condition should be checked at the level `withDifficultyFunc` is called, not the inner wrapper.

fasmat: I strongly disagree. `withDifficultyFunc` returns an option, that is not immediately executed. If it would also return an error a caller of that function e.g. `NewInitializer` would need to check 2 errors per parameter passed in:

current:
```golang
func NewInitializer(opts ...OptionFunc) (*Initializer, error) {
	for _, opt := range opts {
		if err := opt(options); err != nil {
			// if an option returns an error it is detected
			return nil, err
		}
	}
	// ommited for simplicity
}

// Use NewInitializer:
init, err := NewInitializer(WithOption1(5), WithOption2(nil)) // shouldn't panic, since there is an error that can be checked
```

If `withDifficultyFunc` would return an option and an error you need to do double error checking:

```golang
opt1, err := WithOption1(5)
if err != nil {
	// check
}

opt2, err := WithOption2(nil) // why should I expect this to panic?
if err != nil {
	// check
}

init, err := NewInitializer(opt1, opt2) // does error checking again; also shouldn't panic, since there is an error that can be checked
```

Issue #90, opened 2022-12-06
Instead of returning the first Nonce below difficulty, return the absolute lowest


fasmat:
### Summary:

The work oracle is implemented to return the index of the first hash that is below the given difficulty threshold for Pow when calculating labels.

Instead it should return the index of the hash with the lowest value between `StartPosition` and `EndPosition`. Additionally when calculating leaves in batches the `Initializer` should continue to look for indices where the resulting hash is lower than the one already found.

The reason for this is that instead of finding the first Nonce that satisfies the given difficulty, this finds the "best" (lowest) nonce. If a smesher decides to increase their PoST storage in the future this gives them a higher chance of being able to re-use the nonce instead of being required to search for a new one. Additionally if the lowest found nonce doesn't satisfy the difficulty for the larger PoST they can be sure no index in the PoST already calculated does.

### Acceptance criteria:

- add a benchmark to compare the speed of looking for a nonce vs. not using pow during leaf calculation to estimate the impact of the change
  - if looking for a nonce is costly (i.e. speed of initialization drops by more than 10%) we should not implement this change
- when looking for a nonce don't return the first one below the difficulty threshold, but rather the lowest one
- if a nonce was found in a batch during the initialization procedure use the value of the found nonce for the next batch to look for "better" nonces
  - alternatively if it simplifies the implementation `gpu-post` can also just return all nonces that satisfy the difficulty threshold and `post` can use the CPU to find the lowest one among those found, since we are only expecting ~ 8 nonces during initialization the impact of this should be negligible.

### Implementation hints:

- in `gpu-post` the comparison from the calculated hash to `D` (the difficulty threshold) is [here](https://github.com/spacemeshos/gpu-post/blob/5720f29ba2c351ad761ee8b1733cac34d11a7240/src/scrypt-jane/scrypt-jane.cpp#L552-L560). This needs to be changed such that `D` is updated to the value found and the loop isn't aborted.
- in `gpu-post` additionally to the index of the found nonce it's value should be returned as well so it can be used as new difficulty threshold
- in `post` a found nonce should not stop the oracle from looking for one, but instead [here](https://github.com/spacemeshos/post/blob/37b841e69cfffd8774b8a7eeccfe6f7c3a87ac60/initialization/initialization.go#L402-L423) the difficulty should be updated with the found value and the `Initializer` should look for better Nonces in successive batches

noamnelke:
[...]

PR #94, opened 2022-12-12
Replace `opts.NumFiles` with `opts.MaxFileSize`

moshababo:
Part of https://github.com/spacemeshos/go-spacemesh/issues/3529.

Remarks:
* Default `opts.MaxFileSize` value is set to 4 GB.
* Files layout is calculated according to the config and options to accommodate the max size. The last file size might be lower than other files.
* Changing the param value after data was initialized is currently not supported.
* Changing `opts.NumUnits` after data was initialized is supported, and might involve adding new files or removing existing ones.


config/layout.go
```diff
@@ -0,0 +1,27 @@
...
+	NumFiles          uint
+	FileNumLabels     uint64
+	LastFileNumLabels uint64
+}
+
+func DeriveFilesLayout(cfg Config, opts InitOpts) FilesLayout {
+	maxFileSizeBits := opts.MaxFileSize * 8
+	maxFileNumLabels := maxFileSizeBits / uint64(cfg.BitsPerLabel)
+	numLabels := cfg.LabelsPerUnit * uint64(opts.NumUnits)
+	numFiles := numLabels / maxFileNumLabels
```
poszu: I think it will crash if `cfg.MaxFileSize == 0`. Perhaps add a check in `func Validate(cfg Config, opts InitOpts) error` if `cfg.MaxFileSize > 0`?

initialization/initialization.go
```diff
@@ -234,16 +234,28 @@ func (init *Initializer) Initialize(ctx context.Context) error {
 	}
 	defer init.mtx.Unlock()
 
-	numLabels := uint64(init.opts.NumUnits) * uint64(init.cfg.LabelsPerUnit)
-	fileNumLabels := numLabels / uint64(init.opts.NumFiles)
+	init.logger.Info("initialization: datadir: %v, number of units: %v, max file size: %v, number of labels per unit: %v, number of bits per label: %v",
+		init.opts.DataDir, init.opts.NumUnits, init.opts.MaxFileSize, init.cfg.LabelsPerUnit, init.cfg.LabelsPerUnit)
```
poszu, line 238: How about using structured logs here?
```suggestion
	init.logger.With().Info("initialization", log.String("datadir", init.opts.DataDir), (...))
```

initialization/initialization.go
```diff
@@ -298,6 +310,25 @@ func (init *Initializer) Initialize(ctx context.Context) error {
...
 
+func (init *Initializer) removeRedundantFiles(layout config.FilesLayout) error {
+	numFiles, err := init.diskState.NumFiles()
+	if err != nil {
+		return err
+	}
+
+	diff := numFiles - int(layout.NumFiles)
+	for i := 0; i < diff; i++ {
+		index := numFiles - 1 - i
```
poszu: Nit, how about:
```suggestion
	for i := numFiles-1 ; i >= layout.NumFiles; i-- {
```

initialization/initialization_test.go
```diff
@@ -77,6 +76,29 @@ func TestInitialize(t *testing.T) {
 	r.NoError(verifying.VerifyVRFNonce(init.Nonce(), m))
 }
 
+func TestMaxFileSize(t *testing.T) {
+	r := require.New(t)
+
+	cfg := Config{}
+	cfg.BitsPerLabel = 8
+	cfg.LabelsPerUnit = 2048
```
poszu: Nit, how about initializing in-place rather than creating and mutating?
```suggestion
	cfg := Config{
		BitsPerLabel: 8,
		LabelsPerUnit : 2048,
	}
```

config/layout.go
```diff
@@ -0,0 +1,27 @@
+package config
+
+type FilesLayout struct {
+	NumFiles          uint
+	FileNumLabels     uint64
+	LastFileNumLabels uint64
+}
+
+func DeriveFilesLayout(cfg Config, opts InitOpts) FilesLayout {
```
fasmat: Can we move this function / file to the `initialization` package and make `FilesLayout` and `DeriveFilesLayout` private?

I see no need to have this function + type available outside of the `initialization` package

persistence/info.go
```diff
@@ -4,13 +4,13 @@ import (
 	"os"
 )
 
-func NumBytesWritten(dir string, predicate func(os.FileInfo) bool) (uint64, error) {
+func NumBytesWritten(dir string, predicate func(os.FileInfo) bool) (uint64, int, error) {
```
fasmat: This function should be two functions: `NumBytesWritten` and `NumFilesWritten`. There is never a use case where both return values are used, only either the number of bytes or the number of files.

Both functions are also only called in a single place (in the methods of `DiskState`). Imo we can inline these there.

shared/util.go
```diff
@@ -16,8 +16,12 @@ func IsInitFile(file os.FileInfo) bool {
 		return false
 	}
 
+	return IsInitFileName(file.Name())
+}
+
+func IsInitFileName(name string) bool {
```
fasmat: Why extract this as it's own function?

initialization/initialization_test.go
```diff
@@ -751,7 +748,7 @@ func TestValidateMetadata(t *testing.T) {
 		WithLogger(testLogger{t: t}),
 	)
 	r.ErrorAs(err, &errConfigMismatch)
-	r.Equal("NumUnits", errConfigMismatch.Param)
```
fasmat: Is increasing `NumUnits` now allowed? This would possibly require to look for a new VRF nonce.

initialization/initialization_test.go
```diff
@@ -595,9 +600,16 @@ func TestInitialize_MultipleFiles(t *testing.T) {
 			}
 			r.NoError(verifying.VerifyVRFNonce(init.Nonce(), m))
 
-			// TODO(mafa): see above
-			// r.Equal(oneFileNonce, *init.Nonce())
+			if nonce == nil {
```
fasmat: If you handle `numFiles = 1` as a special case for reference. And then increase `numFiles` for every subsequent iteration, this handling of `nonce == nil` becomes unnecessary. The same is true for `data` / `lastData`.

Also this nonce is now never verified to be valid (before it was with: `r.NoError(verifying.VerifyVRFNonce(init.Nonce(), m))`)

initialization/initialization_test.go
```diff
@@ -595,9 +600,16 @@ func TestInitialize_MultipleFiles(t *testing.T) {
...
+				nonce = init.Nonce()
+			} else {
+				// TODO(mafa): since we are not looking for the absolute lowest nonce, we can't guarantee that the nonce will be the same.
+				// see also https://github.com/spacemeshos/post/issues/90
+				// r.Equal(*nonce, *init.Nonce())
```
fasmat: This assert can now probably be enabled again. Changing the number of files for post should not change the found VRF nonce.

initialization/initialization_test.go
```diff
@@ -498,32 +515,37 @@ func TestInitialize_NumUnits_MultipleFiles(t *testing.T) {
...
+	{
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		var eg errgroup.Group
+		eg.Go(assertNumLabelsWritten(ctx, t, init))
+
+		numFiles, err := init.diskState.NumFiles()
+		r.NoError(err)
+		r.Equal(int(opts.NumUnits), numFiles)
```
fasmat: This assertion looks logically incorrect. Why should the number of files equal the number of PoST units?

```suggestion
		layout := config.DeriveFilesLayout(cfg, opts)
		r.Equal(int(layout.NumFiles), numFiles)
```

initialization/initialization_test.go
```diff
@@ -498,32 +515,37 @@ func TestInitialize_NumUnits_MultipleFiles(t *testing.T) {
...
+		r.NoError(init.Initialize(ctx))
+
+		numFiles, err = init.diskState.NumFiles()
+		r.NoError(err)
+		r.Equal(int(opts.NumUnits)-1, numFiles)
```
fasmat: Similarly here:

```suggestion
		newLayout := config.DeriveFilesLayout(cfg, newOpts)
		r.Equal(int(newLayout.NumFiles), numFiles)
		r.Less(int(newLayout.NumFiles), layout.NumFiles)
```

initialization/initialization.go
```diff
@@ -333,6 +364,15 @@ func (init *Initializer) Reset() error {
 	return nil
 }
 
+func (init *Initializer) RemoveFile(name string) error {
+	path := filepath.Join(init.opts.DataDir, name)
+	if err := os.Remove(path); err != nil {
+		return fmt.Errorf("failed to delete file (%v): %v", path, err)
```
fasmat:
```suggestion
		return fmt.Errorf("failed to delete file (%v): %w", path, err)
```

initialization/initialization.go
fasmat: Even simpler imo:

```suggestion
	for i := layout.NumFiles; i < numFiles; i++ {
```

initialization/initialization.go
fasmat: For this we would need to add a hard dependency on `zap`, which I'm not against btw. Instead of `shared.Logger` `WithLogger` should take a `zap.Logger`.

I don't like the `shared.Logger` interface that we currently have as abstraction. It needs an adapter so we can use `zap` and the adapter uses the `SugaredLogger`, which is slower and uses more memory than the non-`Sugared` logger.

The only reason I see for the abstraction is for testing, where we can just use `zaptest.NewLogger`.

initialization/diskstate.go
```diff
@@ -15,14 +15,19 @@ func NewDiskState(datadir string, bitsPerLabel uint) *DiskState {
...
+	numBytesWritten, _, err := d.NumBytesWritten()
 	if err != nil {
 		return 0, err
 	}
 
 	return shared.NumLabels(numBytesWritten, d.bitsPerLabel), nil
 }
 
-func (d *DiskState) NumBytesWritten() (uint64, error) {
+func (d *DiskState) NumBytesWritten() (uint64, int, error) {
```
fasmat: `d.NumBytesWritten` imo should not return the number of files as second parameter.

fasmat (changes requested):

initialization/initialization.go
moshababo: The current logger doesn't support it, so I'd say that it's outside of the scope of this PR.

initialization/initialization.go
moshababo: Yes, wanted to do this but forgot ;) 

Fixed.

initialization/initialization_test.go
moshababo: Fixed.

config/layout.go
moshababo: Fixed.

shared/util.go
moshababo: It's indeed not needed. Fixed.

persistence/info.go
moshababo: Fixed.

initialization/initialization_test.go
moshababo: Do you mean that the new number of labels will invalidate the prior difficulty function?

initialization/initialization_test.go
moshababo: > If you handle `numFiles = 1` as a special case for reference. And then increase `numFiles` for every subsequent iteration, this handling of `nonce == nil` becomes unnecessary. The same is true for `data` / `lastData`.

I made this change because I thought that handling the `nonce == nil` case is much more straightforward than handling the reference case. 

> Also this nonce is now never verified to be valid (before it was with: `r.NoError(verifying.VerifyVRFNonce(init.Nonce(), m))`)

It's still being verified using the same code.

initialization/initialization_test.go
moshababo: It should be the same as before. And it's failing if applied—I guess due to randomness in GPU parallelization.

initialization/initialization_test.go
moshababo: It works because of the actual `cfg` & `opts` values, but I agree that it's confusing. Fixed.

initialization/initialization_test.go
moshababo: https://github.com/spacemeshos/post/pull/94#discussion_r1045900202

initialization/diskstate.go
moshababo: Covered in https://github.com/spacemeshos/post/pull/94#discussion_r1045858080.

initialization/initialization_test.go
fasmat: Yes; because the difficulty depends on `numUnits`.

If we would allow an increase of `numUnits` an adversary could look for a nonce with the smallest PoST allowed (numUnits = 1) and then when they found one that's beneficial for them increase their PoST size to the maximum size allowed.

This would essentially result in the same difficulty for all PoST sizes.

config/layout.go
moshababo: I added `MinMaxFileSize` const and a validation for it.

initialization/initialization_test.go
fasmat: > I made this change because I thought that handling the nonce == nil case is much more straightforward than handling the reference case.

It depends. I think reducing nesting is better for readability than avoiding code duplication:

```golang
var nonce *int
var data []byte

for numFiles := uint(1); numFiles <= 16; numFiles <<= 1 {
	t.Run(fmt.Sprintf("NumFiles=%d", numFiles), func(t *testing.T) {
		if data == nil {
			// only applies to the first loop
			data = lastData
		} else {
			// applies to all other loops
			r.Equal(data, lastData)
		}

		if none == nil {
			// only applies to the first loop
			*nonce = *init.Nonce()
		} else {
			// applies to all other loops
			r.Equal(*nonce, *init.Nonce())
		}
	}
}
```

is much harder to read than:

```golang
// handle numFiles := 1 separately
nonce := init.Nonce()
data := generateData()

for numFiles := uint(2); numFiles <= 16; numFiles <<= 1 {
	t.Run(fmt.Sprintf("NumFiles=%d", numFiles), func(t *testing.T) {
		r.Equal(data, lastData)

		r.Equal(*nonce, *init.Nonce())
	}
}
```

fasmat: > It's still being verified using the same code.

Ah, my bad I misread the diff 😓

initialization/initialization_test.go
fasmat: Then this is a task for after genesis; I already created an issue for it: https://github.com/spacemeshos/post/issues/90

initialization/initialization_test.go
moshababo: I think currently it will just attempt to use a nonce potentially to be found invalid by peers. 

It seems that the solution is to just invalidate both `nonce` and `lastPosition` whenever the number of units is found to be larger than what's stored in the metadata. 
Anything I'm missing?

config/layout.go
fasmat: NIT: `MinFileSize` is probably a less confusing name 😅

initialization/initialization_test.go
moshababo: I still think it's currently simpler to read when considering the code duplication and having value 2 as the starting point.. but I can fix it if you want 😉

initialization/initialization_test.go
fasmat: We have no way of publishing a new nonce when a smesher increases their PoST. This is a ticket for after genesis: https://github.com/spacemeshos/post/issues/90

Increasing PoST size would require recalculating all labels already calculated to look for a new nonce that satisfies the increased difficulty. To avoid this overhead in PoST initialization cost we have to update `gpu-post` to not return the first nonce found to be below the difficulty threashold but the overall lowest nonce it finds in all labels calculated during initialization.

This is the only way we can guarantee that a smesher doesn't have to re-do all the work they already did.

config/layout.go
moshababo: 🤣

initialization/initialization_test.go
fasmat: That would be great ❤️  This would also fix (some of) the issues reported by `staticcheck`.

initialization/initialization_test.go
moshababo: I assumed that it won't require to recalc labels, hence no code changes are needed. The whole point of doing PoW together with the labels is to amortize the work. If labels are already done, you can just continue to do PoW-only for higher positions.

This can indeed introduce a duplicated amount of PoW, and storing the lowest nonce will solve this, but I remembered that @tal-m originally [commented](https://github.com/spacemeshos/pm/issues/172#issuecomment-1245497516) against it. Reading it again, I think we just need to do some benchmarks. 

If other features besides properly amortized PoW are still missing for supporting a different nonce (in ATX publishing, etc.), I guess this is not relevant now. Are you suggesting to disallow the increase of `NumUnits`?

initialization/initialization_test.go
moshababo: Fixed.

initialization/initialization.go
fasmat: I created an issue for this to fix it in a future PR: https://github.com/spacemeshos/post/issues/95

initialization/initialization_test.go
fasmat: Yes, for now I think we should re-add the check that `NumUnits` doesn't increase. We will do this for incoming ATXs: https://github.com/spacemeshos/go-spacemesh/issues/3759 and should also add it here.

After genesis we will update `gpu-post` to return all indices below the difficulty threshold (and selecting for the lowest one in `post`). This will guarantee that no duplicated work is needed when increasing PoST size, after which we will allow increasing `NumUnits`.

fasmat (changes requested):
I think we should re-add the check that `NumUnits` doesn't increase. At least until we have implemented the necessary changes to `gpu-post` to allow for proper amortized work.

moshababo:
It was previously supported, but only limited due to `opts.NumFiles` complications and allowed if `opts.NumFiles=1`. Those complications don't exist with `opts.MaxFileSize`, but I now disallowed it following https://github.com/spacemeshos/post/pull/94#discussion_r1045626345.

fasmat (approved):

fasmat:
LGTM

PR #96, opened 2022-12-16
Add postcli as an additional executable

moshababo:
Adding `postcli`, a CLI tool for PoST initialization, as an additional executable. 

For more info see `README.md`.


cmd/postcli/main.go
```diff
@@ -0,0 +1,181 @@
...
+			log.Panic(err.Error())
+			return
+		}
+		if err == context.Canceled {
+			log.Info("cli: initialization interrupted")
+			return
+		}
+		log.Error("cli: initialization error: %v", err)
+		return
+	}
```
fasmat: `errors.Is` unwraps an error and checks if any (nested) error is of the type, in case the error you are looking for was wrapped:

```suggestion
	err := init.Initialize(ctx)
	switch {
	case errors.Is(err, shared.ErrInitCompleted):
		log.Panic(err.Error())
		return
	case errors.Is(err, context.Canceled):
		log.Info("cli: initialization interrupted")
		return
	case err != nil:
		log.Error("cli: initialization error: %v", err)
		return
	}
```

cmd/postcli/main.go
```diff
@@ -0,0 +1,181 @@
...
+
+	"github.com/davecgh/go-spew/spew"
+
+	"github.com/spacemeshos/ed25519"
+	"github.com/spacemeshos/post/config"
+	"github.com/spacemeshos/post/gpu"
+	"github.com/spacemeshos/post/initialization"
+	"github.com/spacemeshos/post/proving"
+	"github.com/spacemeshos/post/shared"
+	"github.com/spacemeshos/post/verifying"
```
fasmat: To fix this automatically you can install gci and run the formatter yourself:

```
go install github.com/daixiang0/gci@v0.9.0
gci write -s "standard,default,prefix(github.com/spacemeshos/post)" cmd/postcli/
```

```suggestion
	"github.com/davecgh/go-spew/spew"
	"github.com/spacemeshos/ed25519"

	"github.com/spacemeshos/post/config"
	"github.com/spacemeshos/post/gpu"
	"github.com/spacemeshos/post/initialization"
	"github.com/spacemeshos/post/proving"
	"github.com/spacemeshos/post/shared"
	"github.com/spacemeshos/post/verifying"
```

fasmat (approved):

cmd/postcli/main.go
moshababo: Thanks, committed.

Issue #102, opened 2023-02-09
Provide a set of benchmarks / validations that will check the user platform and guesstimate how much storage could be commited


pigmej:
We need to create some "bench" so no one commits too much storage. We need to make sure that:
* benchmark the proof time generation (per space unit)
  * measure disk speed
  * measure CPU speed
  * measure GPU speed for data generation -> that's kinda nice to have part.
* benchmark the proof validation (per proof)

pigmej:
This have to wait for https://github.com/spacemeshos/post/issues/105

pigmej:
@poszu do we need sth more there? What about our current tool? Isn't it https://github.com/spacemeshos/post-rs/issues/34 ?

poszu:
Yes, the benchmark tool is implemented in post-rs. Closing.

Issue #99, opened 2023-01-31
Proof generation and verification is very slow


poszu:
First of all, measuring proof generation performance is not easy as the time to find the proof is random. Changing a single bit of input data might make it do much more work.

I ran some experiments and I was able to come up with input data (8MB POS file) for which it takes 95s to generate a proof on my machine (i7-12700H with 14 cores). Below is the top from pprof:

```
      flat  flat%   sum%        cum   cum%
    40.17s 33.09% 33.09%     40.17s 33.09%  runtime.futex
     8.96s  7.38% 40.48%     36.37s 29.96%  runtime.selectgo
     6.13s  5.05% 45.53%      9.83s  8.10%  runtime.lock2
     6.02s  4.96% 50.49%      6.18s  5.09%  runtime.unlock2
     5.68s  4.68% 55.17%      5.68s  4.68%  github.com/minio/sha256-simd.blockSha
     3.66s  3.02% 58.18%      3.66s  3.02%  runtime.procyield
     2.79s  2.30% 60.48%      7.80s  6.43%  runtime.mallocgc
     2.43s  2.00% 62.48%      2.43s  2.00%  runtime.memmove
     2.25s  1.85% 64.34%      2.76s  2.27%  runtime.casgstatus
     2.15s  1.77% 66.11%     11.38s  9.38%  runtime.sellock
```

The proof generation procedure spends most of the time synchronizing the channel that is used to pass input data to workers label by label (8B at a time).

PR #103, opened 2023-02-10
Speed up proof generation

fasmat:
#### Motivation

Part of #99. This change replaces the current proof generation with a new method using AES that should significantly speed up the process

#### Changes

- A new `proving.Generate` function has been added that implements the new algorithm for proof generation.
- A new `verifying.VerifyNew` function has been added that implements the new algorithm for proof verification.
- The old Prover has been marked as deprecated but not modified/removed to allow for benchmark comparisons
- The parameters for `K1`/`K2`, `N`, `B` have been chosen in a way that they work fast and reliably in unit tests. They are not optimised for security and performance in an actual network. For these parameters see: https://github.com/spacemeshos/go-spacemesh/pull/4071

#### Optional:
- **TODO:** maybe first steps towards #102 (postponed to Rust implementation)


pigmej (commented):
From logic perspective, from speedup perspective yes. But I did not checked the code ;-)

pigmej:
Looked a bit more, @fasmat  what's the reason for keeping the old implementation that we know it's broken?

fasmat:
I already needed it multiple times for reference, like benchmarking and comparison with the current version. I wouldn't have been able to fix the system tests with the new version so quickly if the old version would not have been around any more.

The old algorithms are all marked as deprecated. If any code still uses them we will get a warning in the IDE and in our CI.

We can remove them at any point. Why does it have to be now, when we might still need them for reference?

pigmej:
not only slow, @fasmat it's also broken and missing few items that would make it possible to generate proof correctly.

fasmat:
Again, that's not the point. We are in a transitional period the "new code" is also already decided to be replaced with a new implementation in Rust.

It doesn't hurt to have them around for reference until we have the version we want to use for Genesis.

proving/proving.go
```diff
@@ -3,326 +3,165 @@ package proving
...
-	p.logger = logger
-}
-
-func (p *Prover) verifyGenerateProofAllowed(m *Metadata) error {
-	if err := p.verifyMetadata(m); err != nil {
-		return err
+	select {
+	case <-ctx.Done():
+		return nil, nil, ctx.Err()
+	default:
```
poszu, line 88: Why is this check necessary? Won't the `err ` above (line 81) be `context.Canceled` already?

proving/worker.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	for {
+		data := *(batchDataPool.Get().(*[]byte))
+		n, err := source.Read(data)
+		switch {
+		case err == io.EOF || err == nil:
+			// make sure we don't send partial blocks to the label workers.
+			// TODO(mafa): this silently drops partial blocks and should be handled better.
+			n -= n % int(b)
```
poszu, line 52: I think that's incorrect. The next batch will start with the second part of the labels block. It will offset all subsequent label blocks by `n`.

proving/worker.go
```diff
@@ -0,0 +1,193 @@
...
+			case batchChan <- batch:
+				if err == io.EOF {
+					return nil
+				}
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		default:
+			data = data[:n]
+			batchDataPool.Put(&data)
```
poszu: Why reslice `data` here?

```suggestion
			batchDataPool.Put(&data)
```

proving/worker.go
```diff
@@ -0,0 +1,193 @@
...
+			return err
+		}
+		index += uint64(n) / aes.BlockSize
+	}
+}
+
+// labelWorker is a worker that receives batches from ioWorker and looks for indices to be included in the proof.
+func labelWorker(ctx context.Context, batchChan <-chan *batch, solutionChan chan<- *solution, ch shared.Challenge, numOuts uint8, numNonces uint32, b uint32, d uint, difficulty uint64) error {
+	// use two slices with different types that point to the same memory location.
+	// this is done to speed up the conversation from bytes to uint64.
```
poszu:
```suggestion
	// this is done to speed up the conversion from bytes to uint64.
```

proving/worker.go
```diff
@@ -0,0 +1,193 @@
...
+	out := make([]byte, numOuts*aes.BlockSize+8)
+	u64s := make([][]uint64, 8)
+	for i := range u64s {
+		size := (len(out) - i) / 8
+		u64s[i] = unsafe.Slice((*uint64)(unsafe.Pointer(&out[i])), size)
+	}
```
poszu, line 85: It casts `[]byte` -> `[][]uint64`. For this to be safe, the `out []byte` should probably be 8B-aligned. Are `[]byte` guaranteed to be 8B-aligned in Go?

Casting in the opposite direction seems safer to me.

verifying/verifying.go
```diff
@@ -2,25 +2,17 @@ package verifying
...
 
-type (
-	Config = config.Config
-)
-
-var (
-	WorkOracle = oracle.WorkOracle
-	FastOracle = oracle.FastOracle
-	UInt64LE   = shared.UInt64LE
-)
```
poszu, line 22: I'm curious - why did you remove the re-exports everywhere?

verifying/verifying.go
```diff
@@ -75,34 +79,68 @@ func Verify(p *shared.Proof, m *shared.ProofMetadata) error {
...
+		c, err := oracle.CreateBlockCipher(m.Challenge, nonceBlock+i)
+		if err != nil {
+			return fmt.Errorf("creating cipher for block %d: %w", nonceBlock, err)
+		}
+		ciphers[i] = c
+	}
+
+	block := make([]byte, aes.BlockSize)
+	out := make([]byte, aes.BlockSize*2)
+	u64 := unsafe.Slice((*uint64)(unsafe.Pointer(&out[offset%aes.BlockSize])), 1)
```
poszu: How about creating a pointer to uint64 instead of []uint64 and using it as `*u64` instead of `u64[0]`.
```suggestion
	u64 := (*uint64)(unsafe.Pointer(&out[offset%aes.BlockSize]))
```

verifying/verifying.go
```diff
@@ -75,34 +79,68 @@ func Verify(p *shared.Proof, m *shared.ProofMetadata) error {
...
 	gsReader := shared.NewGranSpecificReader(buf, bitsPerIndex)
-	indicesSet := make(map[uint64]bool, m.K2)
+	indicesSet := make(map[uint64]struct{}, m.K2)
+
+	// create the ciphers for the specific nonce
+	d := shared.CalcD(numLabels, m.B)
+	offset := p.Nonce * uint32(d)
+	nonceBlock := uint8(offset / aes.BlockSize)
+
+	// since the value can be on a boundary between two blocks, we need to create two ciphers
```
poszu, line 99: NIT (opt): it could be predicted if the value is on a boundary or if it fits in a single AES block.

poszu (approved):

proving/proving.go
fasmat: Not necessarily. The `Generate` function receives a context from the caller, which the caller can cancel in case they are not interested in the result any more. This ends all workers, causing `eg.Wait()` to return on line 81, and we need to check here and also return.

Additionally the workers have an `errGroup` context derived from the `Generate` context which they use to signal each other to stop working when they found a solution. If `solutionWorker` found enough indices for a proof it calls `cancel` on the `errGroup` context, causing `ioWorker` and `labelsWorker` to stop and return. This is why `err == context.Cancelled` is not considered an error on line 81, since it may have been caused by the caller of `Generate`, or by `solutionWorker`. This check here is necessary to distinguish between those two cases.

Essentially if `ctx.Done()` doesn't block we don't have a proof, but if it does we have one.

proving/worker.go
fasmat: `Read` can return any number of bytes between `0` and `len(data)`. If it returns fewer than `len(data)` bytes it will also return an non-`nil` error. So the only case this can happen here is if at the end of `source` there are not enough bytes to fill data completely (`err == io.EOF`).

So `n` is something between 0 and `len(data)` and I have to ensure it is a multiple of `B` otherwise we include a few extra bytes in the last AES invocation that have not been read from `source` but are either zeros or leftover from what was in `data` before we passed it to `source.Read`.

There will be no next batch, since after sending this batch the `ioWorker` will return on line 63.

proving/worker.go
fasmat: I'm not sure I can follow. `[]byte` isn't guaranteed to be 8B-aligned, why would that matter in this approach?

Let's say the memory looks like this:

```
offset:  00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F
out:              b0 b1 b2 b3 b4 b5 b6 b7 b8 b9 bA bB bC
u64s[0]:          u0 u0 u0 u0 u0 u0 u0 u0 u1 u1 u1 u1 u1
u64s[1]:             u0 u0 u0 u0 u0 u0 u0 u0 u1 u1 u1 u1
u64s[2]:                u0 u0 u0 u0 u0 u0 u0 u0 u1 u1 u1
...
```

The start of `u64[i]` always points at `out[i]`. I don't know what you mean by casting `uint64` to `byte`? I still need to cast `byte` to `uint64` for all other slices besides the one that is used to cast `[]byte` to?
I am not casting `[]byte` to `[][]uint64` but rather cast each offsets individually to their own separate `[]uint64` (and those are overlapping)

verifying/verifying.go
fasmat: I removed all from the last implementation. I don't quite see the benefit of this? What is the advantage of writing:

```go
res, err := WorkOracle(
    oracle.WithId(...),
    oracle.WithPosition(...),
    ...
)
```

over

```go
res, err := oracle.WorkOracle(
    oracle.WithId(...),
    oracle.WithPosition(...),
    ...
)
```

?

If one calls `WorkOracle` they always have to import `github.com/spacemeshos/post/oracle` anyway, but have 2 ways of calling the same function. `FastOracle` doesn't exist any more and `UInt64LE` isn't used any more.

An argument can be made if one wants to replace these with mocks in tests, but then imho they should be private and we are not using them at the moment this way.

fasmat: These also cause other issues: anyone importing the `verifying` package can change those values during runtime causing all kinds of additional problems 😟

verifying/verifying.go
fasmat: Doing 2 instead of 1 AES calls when it is not needed doesn't impact the performance of verify in a meaningful way. We profiled this function and 99% of the time used to verify a proof is spent in scrypt.

So while it is true that we could optimize this, the gain is negligible and then I'd prefer the more general approach with out handling the special case.

proving/worker.go
fasmat: in the case of `err == nil` `n` is guaranteed to be a multiple of `b` so there will only ever be labels dropped at the end of `source` and at most `b - 1`.

poszu: 👍

proving/worker.go
poszu: I'm not sure about Go. Although in C/C++ it matters. A pointer to uint64 should be aligned to 8B in order for the CPU to load it efficiently into registers. Otherwise, it could do it inefficiently or even crash (https://stackoverflow.com/questions/1496848/does-unaligned-memory-access-always-cause-bus-errors/1496881#1496881).


By
> casts `[]byte` -> `[][]uint64`
I meant that you reinterpret the memory under `[]byte` as `[][]uint64` (so that the u64 slices point to the underlying byte array).

I checked on my PC and accessing bytes from `out` as `uint64` in an unaligned way _works_ so it probably is fine on x86.

fasmat: It causes a performance hit. According to profiled benchmarks the proof generator spends like 25% on Line 117:

```
val := u64s[offset%8][offset/8] & mask // mask out the bits we don't care about
```

compared to 65% on `cipher[i].Encrypt`.

It is slow but still a bit faster than bit masking and shifting the bits of interest. Regarding unaligned access, there was an issue in go on `mips64` that has been fixed: https://github.com/golang/go/issues/46283

I couldn't find any information in the spec if unaligned access is problematic.

PR #109, opened 2023-03-03
Use proof gen from postrs

### LINKED ISSUE #105: Rewrite the proving algorithm in rust
The current implementation is limited by Golang possibilities https://github.com/spacemeshos/post/issues/99
We have done PoC version in Rust.

We need to follow with full rust implementation of the algorithm and then integration of it to the Golang codebase.

Besides the implementation we need to make sure that we remember about all the needed improvements that is (but not limited to):
* if proof not found then iterate again with another nonces
* invalid disk reads https://github.com/spacemeshos/post/issues/101

poszu:
Use the Rust implementation for proof generation.

Closes #105


proving/postrs/bridge.go
```diff
@@ -0,0 +1,56 @@
+package postrs
+
+// #cgo LDFLAGS: -L${SRCDIR} -lpost
+
```
poszu:
```suggestion
```

Makefile.Inc
```diff
@@ -70,12 +70,44 @@ $(BINDIR_GPU_SETUP_LIBS): $(PROJ_DIR)$(GPU_SETUP_ZIP)
...
+  POSTRS_SETUP_ZIP = libpost-windows-latest.zip
+else
+  ifeq ($(platform), $(filter $(platform), macos macos-m1))
+    POSTRS_SETUP_LIBS = prover.h libpost.dylib
+	POSTRS_SETUP_ZIP = libpost-macos-latest.zip
+  else
+    POSTRS_SETUP_LIBS = prover.h libpost.so
+	POSTRS_SETUP_ZIP = libpost-ubuntu-20.04.zip
+  endif
+endif
```
fasmat: For GPU post we are targeting ubuntu 22.04.

Also is there a `linux arm64` and `macos m1` build of the library?

proving/postrs/.gitignore
```diff
@@ -0,0 +1,3 @@
+prover.h
+libpost.a
+post.lib
```
fasmat: Not needed if the files are in the `build` directory (see other comment).

proving/postrs/bridge.go
```diff
@@ -0,0 +1,54 @@
+package postrs
+
+// #cgo LDFLAGS: -L/home/bartosz/workspace/post-rs/target/release-clib -lpost
```
fasmat: You have hardcoded your home directory here 😉 

The Makefile should already set the `build` directory as a lookup path for the library, so if you put it there this should be enough:

```suggestion
// #cgo LDFLAGS: -lpost
```

proving/postrs/bridge.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"unsafe"
+
+	"github.com/spacemeshos/post/config"
+)
+
+type Proof struct {
+	Nonce    uint32
+	Indicies []uint64
+}
```
fasmat: Why not use `shared.Proof` instead of redefining the type here?

verifying/verifying_test.go
```diff
@@ -18,6 +18,7 @@ import (
 
 func getTestConfig(tb testing.TB) (config.Config, config.InitOpts) {
 	cfg := config.DefaultConfig()
+	cfg.LabelsPerUnit = 4096
```
fasmat: If tests don't work with the default config, imo we should make `4096` the new default.

proving/proving_test.go
```diff
@@ -218,104 +214,3 @@ func Test_Generate_TestNetSettings(t *testing.T) {
...
-		b.StartTimer()
-		start := time.Now()
-		proof, proofMetadata, err := Generate(context.Background(), ch, cfg, &shared.DisabledLogger{}, WithDataSource(cfg, nodeId, commitmentAtxId, opts.DataDir))
-		r.NoError(err)
-		b.ReportMetric(time.Since(start).Seconds(), "sec/proof")
-		b.StopTimer()
-
-		r.NoError(verifying.Verify(proof, proofMetadata))
-	}
-}
```
fasmat, line 321: Please leave this test (or update it if necessary) so we can check the performance of Generating / Verifying proofs with systest settings.

If the current systest settings don't work with the new implementation they have to be updated as well... (in `fastnet.go` in the `go-spacemesh` repository).

The default settings are the ones used for testnet, not for system tests.

proving/proving_test.go
```diff
@@ -21,6 +16,7 @@ import (
 
 func getTestConfig(tb testing.TB) (config.Config, config.InitOpts) {
 	cfg := config.DefaultConfig()
+	cfg.LabelsPerUnit = 4096
```
fasmat: Same as the other comment, if the current default config doesn't work, this should be the new default config instead.

proving/postrs/bridge.go
```diff
@@ -0,0 +1,54 @@
...
+
+func GenerateProof(datadir string, challenge []byte, cfg config.Config) (*Proof, error) {
+	datadirPtr := C.CString(datadir)
+	defer C.free(unsafe.Pointer(datadirPtr))
+
+	challengePtr := C.CBytes(challenge)
+	defer C.free(challengePtr)
+
+	config := C.Config{
+		n:               20,
```
fasmat: why not `cfg.N`?

proving/postrs/bridge.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type Proof struct {
+	Nonce    uint32
+	Indicies []uint64
+}
+
+func GenerateProof(datadir string, challenge []byte, cfg config.Config) (*Proof, error) {
+	datadirPtr := C.CString(datadir)
+	defer C.free(unsafe.Pointer(datadirPtr))
```
fasmat, line 22: does this not work?

```suggestion
	defer C.free(datadirPtr)
```

proving/postrs/bridge.go
```diff
@@ -0,0 +1,54 @@
...
+		config,
+	)
+
+	if cProof == nil {
+		return nil, fmt.Errorf("got nil")
+	}
+	defer C.free_proof(cProof)
+
+	indicies := unsafe.Slice((*uint64)(unsafe.Pointer(cProof.indicies.ptr)), cProof.indicies.len)
+	indicies = append([]uint64{}, indicies...)
```
fasmat: This seems confusing to me. How about this:

```suggestion
	indicies := make([]uint64, cProof.indicies.len)
	copy(indicies, unsafe.Slice((*uint64)(unsafe.Pointer(cProof.indicies.ptr)), cProof.indicies.len))
```

fasmat (commented):
Good work! I added a few comments on some minor things.

Is a macos M1 / linux arm64 version of postrs available? Otherwise I cannot run it on my machine.

proving/postrs/bridge.go
fasmat: Is the explicit copy even necessary? There is `C.GoBytes` that converts a pointer to a byte array to a `[]bytes` maybe something similar exists for `uint64` slices/arrays?

proving/postrs/bridge.go
poszu: I know 😬 It's already removed.

proving/proving_test.go
poszu: The problem is that the library reads from the files on the disk and runs until a proof is found. There is no way to make it "process N GiB of data and return" to be able to benchmark.

The proper benches measuring raw speed are in post-rs now.

proving/postrs/bridge.go
poszu: I copy it over to Go to be able to free the memory allocated on the Rust side.

Edit: OK, so it turns out that `C.GoBytes` also copies the memory "into Go"

proving/postrs/bridge.go
poszu: No:
>cannot use datadirPtr (variable of type *_Ctype_char) as unsafe.Pointer value in argument to C.free

proving/postrs/bridge.go
poszu: It's a different type. It contains indices as uint64s that are later compressed. That compression is not implemented on Rust side (I'm not even sure if it's worth implementing there, at least now).

Makefile.Inc
poszu: Actually, gpu-post builds library on ubuntu 20.04: https://github.com/spacemeshos/gpu-post/blob/949f87501ba904504b274bab7e0ff63d01b4aa8d/.github/workflows/build.yml#L17

post-rs library is built on ubuntu-20.04 and ubuntu-latest runners.

poszu: Here are the platforms the library is built on: https://github.com/spacemeshos/post-rs/blob/3af14007db623c5b5be5c16b9191032d9d6532c6/.github/workflows/ci.yml#L80-L94

fasmat: > Actually, gpu-post builds library on ubuntu 20.04: https://github.com/spacemeshos/gpu-post/blob/949f87501ba904504b274bab7e0ff63d01b4aa8d/.github/workflows/build.yml#L17
> 
> post-rs library is built on ubuntu-20.04 and ubuntu-latest runners.

That's weird I could have sworn we are using Ubuntu 22.04, maybe just for linux arm64....

proving/postrs/bridge.go
fasmat: Since this type is only required internally we could probably make it private (the Public fields are still accessible from other packages). Just to avoid confusion between `shared.Proof` and `postrs.Proof`.

Makefile.Inc
poszu: PR adding building the library for arm64: https://github.com/spacemeshos/post-rs/pull/7

proving/postrs/bridge.go
fasmat: Copying it is fine and correct; I just advocate to use `copy` over `append` to make it clear that this is a copy 🙂

fasmat (approved):

fasmat:
lgtm

PR #111, opened 2023-03-10
Integrate proof verification implemented in post-rs

### LINKED ISSUE #99: Proof generation and verification is very slow

### LINKED ISSUE #104: Add the POW to the AES invocations so running proving with multiple nonces is not that cheap.
#### Description

We need to make using multiple (100s of thousands) nonces in proving harder. Therefore we need to add POW to the initial phase of AES.

#### Acceptance criteria

- Define POW algorithm together with the research team
   - The goal is to prevent an adversary to grind on challenges for keys that would allow them to generate a valid PoST proof with without having the labels stored on disk.
- Limit maximum nonce for a challenge to 40
   - This does not mean that the reference implementation of PoST uses 40 nonces to look for indices that satisfy the difficulty threshold. Just that when verifying a proof by another identity nonces >= 40 will be rendered invalid

### LINKED ISSUE #217: Adjust POST algorithm to the latest agreements
We need to adjust POST algorithm to the recent agreements:
* `K2=8000`
* `K3=50`

`K3` is the subset of `K2` that is used to validate the proof. We need to make sure that each node will always validate the same `K3` indices from given K2 indices. We should use Fisher-Yates shuffle.

Besides that we should increase the bits per label to `16` so also we need to change `N` to script to `8192`.

poszu:
Few changes in this PR:
- label size was changed from 1B to 16B. Scrypt difficulty (`N`) was scaled up 16x to balance it out (so that initialization speed in terms of B/s stays the same),
- use blake3 to hash `node_id` and `commitment_atx_id` together into _commitment_ for initialization,
- integrated proof verification implemented in https://github.com/spacemeshos/post-rs,
- There are new parameters:
  * K3 - the size of subset of proof indices that actually verified
  * K2 Proof of Work difficulty - to prevent grinding on nonces when generating a proof. It adds cost to trying another nonce. Thanks to this, trying many nonces is not cost-effective.
  * K3 Proof of Work difficulty
 - Proof structure now contains two new values - K2 and K3 PoW.

Closes #99 
Closes #104 
Closes https://github.com/spacemeshos/pm/issues/217


config/config.go
```diff
@@ -57,18 +61,23 @@ type Config struct {
 	B  uint32 // B is the number of labels used per AES invocation when generating a proof. Lower values speed up verification, higher values proof generation.
 	N  uint32 // N is the number of nonces tried at the same time when generating a proof.
```
fasmat: This can probably be removed from `Config` and for now be made a constant.

Later it should be made an option that can be configured by the node.

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/spacemeshos/sha256-simd v0.1.0
 	github.com/stretchr/testify v1.8.2
 	golang.org/x/sync v0.1.0
+	lukechampine.com/blake3 v1.1.7
```
fasmat: Why not `github.com/zeebo/blake3`? That's the library we use in `go-spacemesh`

oracle/oracle.go
```diff
@@ -169,20 +171,22 @@ func WorkOracle(opts ...OptionFunc) (WorkOracleResult, error) {
...
+	hasher := blake3.New(16, nil)
+	hasher.Write(ch)
+
+	nonceBytes := make([]byte, 4)
+	binary.LittleEndian.PutUint32(nonceBytes, nonceGroup)
+	hasher.Write(nonceBytes)
+
+	k2PowBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(k2PowBytes, k2Pow)
+	hasher.Write(k2PowBytes)
```
fasmat: This is shorter and saves on allocating temporary memory:

```suggestion
	binary.Write(hasher, binary.LittleEndian, ch)
	binary.Write(hasher, binary.LittleEndian, nonceGroup)
	binary.Write(hasher, binary.LittleEndian, k2Pow)
```

proving/proving_test.go
```diff
@@ -16,6 +16,8 @@ import (
 
 func getTestConfig(tb testing.TB) (config.Config, config.InitOpts) {
 	cfg := config.DefaultConfig()
+	cfg.K2PowDifficulty = 0x00FFFFFF_FFFFFFFF
+	cfg.K3PowDifficulty = 0x00FFFFFF_FFFFFFFF
```
fasmat: Could we make these the default values instead of `math.MaxUint64`. Maybe `0x0FFFFFFF_FFFFFFFF` is a better default, since on average a node would only need to try 16 values but then we would also have this in system tests, otherwise the systemtests would not do PoW on k2 and k3

verifying/verifying.go
```diff
@@ -91,23 +90,15 @@ func Verify(p *shared.Proof, m *shared.ProofMetadata, opts ...OptionFunc) error
...
-	// since the value can be on a boundary between two blocks, we need to create two ciphers
-	ciphers := make([]cipher.Block, 2)
-	for i := uint8(0); i < 2; i++ {
-		c, err := oracle.CreateBlockCipher(m.Challenge, nonceBlock+i)
-		if err != nil {
-			return fmt.Errorf("creating cipher for block %d: %w", nonceBlock, err)
-		}
-		ciphers[i] = c
+	nonceBlock := p.Nonce / 2
+	cipher, err := oracle.CreateBlockCipher(m.Challenge, nonceBlock, p.K2Pow)
```
fasmat: `p.K2Pow` needs to be validated to be a valid PoW for the challenge and nonce.

go.mod
poszu: I just picked the first one I found. I will switch to zeebo for consistency. Thanks!

verifying/verifying.go
poszu: I know. I'm not implementing the verification part yet. For now, I just updated the AES cipher key calc so that verification works.

poszu: Verification's been ported to post-rs

config/config.go
```diff
@@ -74,8 +71,6 @@ func DefaultConfig() Config {
 		MinNumUnits:     DefaultMinNumUnits,
 		K1:              DefaultK1,
 		K2:              DefaultK2,
-		B:               DefaultAESBatchSize,
```
fasmat: `DefaultK3` is missing 😉

config/config.go
```diff
@@ -87,6 +82,40 @@ type InitOpts struct {
...
+	if p.R == 0 {
+		return errors.New("scrypt parameter r cannot be 0")
+	}
+	if p.P == 0 {
+		return errors.New("scrypt parameter p cannot be 0")
+	}
+	return nil
+}
+
+func DefaultPowScryptParams() ScryptParams {
```
fasmat: NIT `DefaultPowParams` and `DefaultLabelParams` is shorter and makes confusing the two less likely. Scrypt is already in the type.

config/config.go
```diff
@@ -100,6 +129,7 @@ func DefaultInitOpts() InitOpts {
 		MaxFileSize:       DefaultMaxFileSize,
 		ComputeProviderID: BestProviderID,
 		Throttle:          false,
+		Scrypt:            DefaultLabelsScryptParams(),
```
fasmat: NIT: same here, instead of `Scrypt` I would call it `LabelParams` or `InitParams`.

gpu/api.go
```diff
@@ -189,12 +200,18 @@ func WithComputePow(difficulty []byte) OptionFunc {
...
+		return nil
+	}
+}
+
 // ScryptPositions computes the scrypt output for the given options.
 func ScryptPositions(opts ...OptionFunc) (*ScryptPositionsResult, error) {
 	options := &option{
-		n:             512,
-		r:             1,
-		p:             1,
```
fasmat, line 197: I like the `WithScryptParams` functional option, but I would still leave default values for these parameters.

It will always be `WithScryptParams(8192, 1,1)` outside of tests. I think we keep default values (and only change `n` from 512 to 8192)

gpu/api_test.go
```diff
@@ -304,13 +314,14 @@ func Test_ScryptPositions_Pow(t *testing.T) {
 				WithComputeProviderID(p.ID),
 				WithCommitment(commitment),
 				WithSalt(salt),
-				WithStartAndEndPosition(0, 128*1024),
-				WithBitsPerLabel(8),
+				WithStartAndEndPosition(0, 256*1024),
+				WithBitsPerLabel(8*16),
```
fasmat, line 318: Would it make sense to change this parameter to `WithBytesPerLabel` and stop supporting label sizes that are not full bytes?

oracle/oracle.go
```diff
@@ -125,6 +126,13 @@ func WithComputePow(difficulty []byte) OptionFunc {
 	}
 }
 
+func WithScryptParams(params config.ScryptParams) OptionFunc {
+	return func(opts *option) error {
+		opts.scrypt = &params
+		return nil
+	}
+}
```
fasmat, line 134: NIT: same as above - I think it makes sense to not have to set this explicitly and fall back to the default N=8192, r=1, p=1 if not set.

I'd prefer explicit N, r and p parameters, otherwise one can call `WithScryptParams(config.ScryptParams{})` which is not valid and only checked for in `gpu.ScryptPostitions`.

proving/proving.go
```diff
@@ -51,29 +32,22 @@ func Generate(ctx context.Context, ch shared.Challenge, cfg config.Config, logge
...
 	proofMetadata := &shared.ProofMetadata{
 		NodeId:          options.nodeId,
 		CommitmentAtxId: options.commitmentAtxId,
 		Challenge:       ch,
-		BitsPerLabel:    cfg.BitsPerLabel,
 		LabelsPerUnit:   cfg.LabelsPerUnit,
 		NumUnits:        options.numUnits,
 		K1:              cfg.K1,
 		K2:              cfg.K2,
-		N:               cfg.N,
```
fasmat, line 75: does `K3` need to be part of the Metadata?

proving/proving.go
```diff
@@ -51,29 +32,22 @@ func Generate(ctx context.Context, ch shared.Challenge, cfg config.Config, logge
...
 	logger.Info("proving: generated proof")
-	logger.Debug("Nonce: %v, Indicies: %v, K2PoW: %d, K3PoW: %d", result.Nonce, result.Indicies, result.K2Pow, result.K3Pow)
-	indicies, err := compressIndicies(result.Indicies, options.numUnits, cfg)
-	if err != nil {
-		return nil, nil, fmt.Errorf("compressing proof indicies: %w", err)
-	}
+	logger.Debug("Nonce: %v, Indices: %v, K2PoW: %d, K3PoW: %d", result.Nonce, result.Indices, result.K2Pow, result.K3Pow)
 
-	proof := &shared.Proof{Nonce: result.Nonce, Indices: indicies, K2Pow: result.K2Pow, K3Pow: result.K3Pow}
+	proof := &shared.Proof{Nonce: result.Nonce, Indices: result.Indices, K2Pow: result.K2Pow, K3Pow: result.K3Pow}
```
fasmat, line 43: Indices are not compressed any more? That should make ATXs a lot bigger.

If a node has 1 TB of PoST data, only 40 bits are needed to encode label indices, but the indices are uint64s. So the 24 MSB are always 0. Even 1 PB has 14 bits that are always 0.

ATXs are part of the mesh, so if we can save space there we should imo. We can still do the compression and decompression in go, since it should be much faster than the actual proof generation & verification.

proving/proving_test.go
```diff
@@ -66,28 +65,37 @@ func Test_Generate(t *testing.T) {
...
 			r.Equal(len(ch), n)
 
-			proof, proofMetaData, err := Generate(context.Background(), ch, cfg, log, WithDataSource(cfg, nodeId, commitmentAtxId, opts.DataDir))
+			proof, proofMetaData, err := Generate(
+				context.Background(),
+				ch,
+				cfg,
+				log,
+				WithDataSource(cfg, nodeId, commitmentAtxId, opts.DataDir),
+				WithPowScryptParams(config.DefaultPowScryptParams()))
```
fasmat: Super-NIT 😄 : I don't know why gofmt doesn't do this automatically:

```suggestion
				WithPowScryptParams(config.DefaultPowScryptParams()),
			)
```

imo leads to better visual signalling to what is part of the parameter list and if you add or remove a parameter only one line shows up in the diff (instead of two if you add or remove one at the end of the list).

verifying/verifying.go
```diff
@@ -50,19 +57,18 @@ func VerifyVRFNonce(nonce *uint64, m *shared.VRFNonceMetadata) error {
...
 	for _, opt := range opts {
-		opt(options)
+		if err := opt(options); err != nil {
+			return err
+		}
 	}
-	if err := options.validate(); err != nil {
-		return err
+	if m.K1 != cfg.K1 {
+		return fmt.Errorf("invalid `K1` in proof metadata: %d != %d", m.K1, cfg.K1)
```
fasmat: I don't think this should be checked for here. `Verify` verifies that the given `Proof` is valid with the given `ProofMetadata`.

The validation of the metadata at the moment happens in `go-spacemesh`. Alternatively we could have a `VerifyMetadata(m *shared.ProofMetadata, cfg config.Config)`. That does these checks.

verifying/verifying.go
```diff
@@ -50,19 +57,18 @@ func VerifyVRFNonce(nonce *uint64, m *shared.VRFNonceMetadata) error {
 
 // Verify ensures the validity of a proof in respect to its metadata.
 // It returns nil if the proof is valid or an error describing the failure, otherwise.
-func Verify(p *shared.Proof, m *shared.ProofMetadata, opts ...OptionFunc) error {
-	options := &option{
-		logger: &shared.DisabledLogger{},
-	}
+func Verify(p *shared.Proof, m *shared.ProofMetadata, cfg config.Config, opts ...OptionFunc) error {
```
fasmat, line 60: small rant unrelated to the PR (but maybe worth a ticket):

This is the reason why I think `config` is a bad package name. What is `config.Config` for? The PoST config? The PoST Metadata config? The Node config? The poet config? All of these are referred to as `config.Config` 😅 

Better is: `config.Post`, `config.Node`, etc. but then one still has issues with colliding package names.

My preferred naming scheme would be: `post.Config` or `post.MetaDataConfig`, `node.Config`, `poet.Config`.

This would mean that packages like `config` and `shared` should be together in a `post` package, which should be the base package for all packages using them:

`post` (contains current `config` & `shared` content)
- `post/proof` (contains `Generate` and `Verify` functions and types related with them)
- `post/init` (for initialisation - possibly even better to have it as part of the base `post` package)
- `post/oracle` (but instead of `oracle.WorkOracle` -> `oracle.CalcLabels`)
- other packages for C and rust bridges (named like `internal/postrs`, `internal/gpu`, etc. so they cannot be imported from outside the post module)

gpu/api.go
poszu: I think that the C-library wrapper shouldn't define the defaults and require params to be explicitly set.

gpu/api_test.go
poszu: Perhaps 🤷 The gpu-post supports it so it seems to me unnecessary to artificially limit it.

proving/proving.go
poszu: I don't know this. I'd argue that neither K1 nor K2 nor K3 needs to be there.

proving/proving.go
poszu: Indices are compressed in post-rs already.

verifying/verifying.go
poszu: Hmm, IMO it's wrong that metadata is validated separately in go-spacemesh. I think it should be moved here to keep proof verification consistent in a single place.

config/config.go
poszu: I'm not sure about this one. The structure gathers label initialization options. This information is already included in its name. It's sort of obvious that the options that it contains control label initialization. The field name `Scrypt` informs the reader that this field controls the scripting part of it.

proving/proving.go
poszu: Actually, definitely no - K3 is parameter influencing verification only

verifying/verifying_test.go
```diff
@@ -112,80 +124,59 @@ func TestVerifyPow(t *testing.T) {
...
 func Benchmark_Verify_Fastnet(b *testing.B) {
 	r := require.New(b)
-
 	nodeId := make([]byte, 32)
 	commitmentAtxId := make([]byte, 32)
 	ch := make(shared.Challenge, 32)
 
 	cfg, opts := getTestConfig(b)
-	cfg.BitsPerLabel = 8
+	cfg.LabelsPerUnit = 32
```
fasmat: This property is specified twice in this test

config/config.go
```diff
@@ -87,10 +108,11 @@ const BestProviderID = -1
 func DefaultInitOpts() InitOpts {
 	return InitOpts{
 		DataDir:           DefaultDataDir,
-		NumUnits:          DefaultMinNumUnits + 1,
-		MaxFileSize:       DefaultMaxFileSize,
+		NumUnits:          defaultMinNumUnits + 1,
+		MaxFileSize:       4 * 1024 * 1024 * 1024, // 4 GB,
```
fasmat: I would prefer the constant over `4 * 1024 * 1024 * 1024`. You can even do something like:

```golang
type ByteSize uint64

const (
	B   ByteSize = 1
	KiB ByteSize = 1 << (10 * iota)
	MiB
	GiB
	TiB
	PiB
	EiB
)

const defaultMaxFileSize = 4 * GiB
```

verifying/verifying.go
fasmat: I still would prefer that `Verify` only checks if the given proof is valid with the given metadata.

To check the metadata is valid, we could add a `VerifyMetadata(m *shared.ProofMetadata, cfg config.Config)` function that then replaces the checks currently present in `go-spacemesh`.

verifying/verifying_options.go
```diff
@@ -1,31 +1,46 @@
...
-import "github.com/spacemeshos/post/shared"
+import (
+	"github.com/spacemeshos/post/config"
+	"github.com/spacemeshos/post/shared"
+)
 
 type option struct {
-	verifyFunc func(val uint64) bool
-
 	logger shared.Logger
```
fasmat: I think this isn't used anywhere any more.

verifying/verifying_options.go
```diff
@@ -1,31 +1,46 @@
...
+		labelScrypt: config.DefaultLabelParams(),
+	}
 }
 
 type OptionFunc func(*option) error
 
-// withVerifyFunc sets a custom verify function. This is provided for testing purposes, and should not be used in production.
-func withVerifyFunc(f func(val uint64) bool) OptionFunc {
+// WithLogger adds a logger to the verifier to log debug messages.
+func WithLogger(logger shared.Logger) OptionFunc {
```
fasmat: The logger isn't used anymore and can be removed.

verifying/verifying_options.go
```diff
@@ -1,31 +1,46 @@
...
 	logger shared.Logger
+	// scrypt parameters for AES PoW
+	powScrypt config.ScryptParams
+	// scrypt parameters for labels initialization
+	labelScrypt config.ScryptParams
 }
 
-func (o *option) validate() error {
-	return nil
+func defaultOpts() *option {
```
fasmat, line 14: I think this is only used in a single place and can thereby just be inlined.

proving/proving_options.go
```diff
@@ -49,3 +58,27 @@ func WithDataSource(cfg config.Config, nodeId, commitmentAtxId []byte, datadir s
 		return nil
 	}
 }
+
+func defaultOpts() *option {
```
fasmat: I think this is only used in a single place and can therefore be inlined.

shared/proof.go
```diff
@@ -11,14 +13,10 @@ type ProofMetadata struct {
 
 	Challenge     Challenge
 	NumUnits      uint32
-	BitsPerLabel  uint8
 	LabelsPerUnit uint64
 
 	K1 uint32 // K1 specifies the difficulty for a label to be a candidate for a proof.
 	K2 uint32 // K2 is the number of labels below the required difficulty required for a proof.
```
fasmat: Does `K3` need to be added to this?

proving/proving_test.go
```diff
@@ -166,14 +164,14 @@ func Test_Generate_TestNetSettings(t *testing.T) {
...
 	cfg := config.DefaultConfig()
 
-	// https://colab.research.google.com/github/spacemeshos/notebooks/blob/main/post-proof-params.ipynb
-	cfg.LabelsPerUnit = 2 << 16
-	cfg.B = 16
-	cfg.K1 = 279
-	cfg.K2 = 287
-	cfg.N = 24
+	// Test-net settings:
+	cfg.LabelsPerUnit = 20 * 1024 / 16 // 20kB unit
```
fasmat: With the suggestion above this could be:

```suggestion
	cfg.LabelsPerUnit = 20 * KiB / (16*B) // 20kB unit
```

shared/proof.go
poszu: No, `K3` is a verification parameter unrelated to the proof itself (it isn't used in proof generation). `K3` defines the size of a subset of the proof's labels that the verifier must check.

verifying/verifying_options.go
poszu: It's actually used in two places.

proving/proving_options.go
poszu: What's wrong with defining the defaults next to structure definition? I think it makes it easier to find.

config/config.go
poszu: Done

verifying/verifying.go
poszu: This function already checks some parts of metadata (`NodeId` and `CommitmentAtxId`) - introduced in #67. 

Why would you prefer it this way? It would indicate that one can verify the proof against multiple metadata.

IMO there is no point in verifying proof and metadata separately. _The proof_ consists of a pair: `Proof` and `ProofMetadata`. Both must be correct at the same time. In fact, I don't really understand why they are separated in the first place (`NiPost` contains both)...

verifying/verifying.go
poszu: I agree. `config.Config` is confusing. Would you mind creating an issue and explaining the approach you propose there?

config/config.go
fasmat: I mean `DefaultLabelParams()` instead of `DefaultLabelsScryptParams()`

verifying/verifying.go
fasmat: I feel like it would be cleaner to validate the metadata and the proof separately. Imo the length checks for `NodeId` and `CommitmentAtx` should also be part of the `VerifyMetadata` function instead of this function.

I'm not insisting on this change I just think it would be cleaner

proving/proving_options.go
fasmat: It's another layer of indirection that every time we read the code needs extra navigation:

`GenerateProof` -> `defaultOpts()` -> `config.DefaultPowParams()` instead of
`GenerateProof` -> `config.DefaultPowParams()`

It's a function that imo only exists to name these as the default options, which is already clear from the fact that the `option` struct is initialized with them and then iterated over with the functional options 🤷

verifying/verifying_options.go
fasmat: The same applies here as I've written to the other `defaultOpts` I don't see the benefit of marking these explicitly as default options, when this would be just as clear from:

```golang
opts := &options{
    // defaults
}
for _, opt := range optFuncs {
    opt(opts)
}
```

Its an additional indirection, that I don't see the benefit of 🤷

proving/proving.go
fasmat: Depends on how you see it. A proof is valid if all K3 indices pass the difficulty check even if some of the K2 labels do not pass it, so a proof can be valid with one K3 parameter and invalid with a different one.

shared/proof.go
fasmat: But a proof is valid if the `K3` labels checked are valid, not if all labels are valid. So different `K3` values can cause a proof to be valid or not...

postrs/bridge.go
```diff
@@ -0,0 +1,138 @@
...
+	if len(proof.Indices) == 0 {
+		return errors.New("proof indices are empty")
+	}
+
+	config := C.Config{
+		k1:                C.uint32_t(cfg.K1),
+		k2:                C.uint32_t(cfg.K2),
+		k3:                C.uint32_t(cfg.K3),
+		k2_pow_difficulty: C.uint64_t(cfg.K2PowDifficulty),
+		k3_pow_difficulty: C.uint64_t(cfg.K3PowDifficulty),
```
fasmat, line 97: Why the values from `cfg` and not from `metadata`?

verifying/verifying_options.go
poszu: The benefit is that defaults are defined in a single place and cannot go out of sync in contrast to if they were defined twice.

postrs/bridge.go
poszu: Because K3 and pow difficulties are not part of metadata. For K1 and K2 it doesn't make any difference as they must be equal.

shared/proof.go
poszu: `K3` is not used for proof generation at all. It's a parameter solely for a verifier. The proof is not created "with some K3". K3 could probably even change long after proof had been generated and it should still be valid (unless the prover had ground and had found some subset of labels that satisfy a particular value of K3).

proving/proving_options.go
poszu: Removed `defaultOpts()` for proving options but I left it for verifying options to avoid getting defaults out-of-sync.

fasmat (approved):
LGTM

verifying/verifying.go
fasmat: added issue here: https://github.com/spacemeshos/post/issues/116

PR #114, opened 2023-03-20
Remove gran-specific reader and writer

poszu:


initialization/initialization_test.go
```diff
@@ -878,20 +877,7 @@ func initData(datadir string, bitsPerLabel uint) ([]byte, error) {
...
-			if err == io.EOF {
-				break
-			}
-			return nil, err
-		}
-		writer.Write(b)
-	}
-
-	return writer.Bytes(), nil
+	return io.ReadAll(reader)
```
fasmat, line 859: It's still the same behaviour as before so I guess it's OK, but `initData` reads ALL the data from disk into a single `[]byte` and returns it.

I assume this will cause issues when `datadir` contains more than a few MiB of data... Since this is only used in tests it's probably OK though.

fasmat (approved):

initialization/initialization_test.go
poszu: Yes - it's only used in tests where the data is small.

PR #119, opened 2023-03-31
Writer shouldn't seek on every write

fasmat:


piersy (approved):
This looks fine to me, although there seems to be a change of crypto library mixed in as well.

cmd/postcli/main.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"crypto/ed25519"
```
piersy, line 5: Whats with the change here?

fasmat: Our `github.com/spacemeshos/ed25519` library has been archived because we are not using key extraction any more.

We were only using `ed25519.GenerateKey` which is the same code in both libraries, so this change is a `no-op` and gets rid of an dependency we are not maintaining any more.

PR #126, opened 2023-04-18
Add validation for compute batch size

piersy:
Previously compute batch size was not being validated this could lead to crashes if an invalid value was used.

See - https://github.com/spacemeshos/go-spacemesh/issues/4287

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -12,8 +12,6 @@ import (
 const (
 	DefaultDataDirName = "data"
 
-	// DefaultComputeBatchSize value must be divisible by 8, to guarantee that writing to disk
-	// and file truncating is byte-granular.
 	DefaultComputeBatchSize = 1 << 20
 
 	MinBitsPerLabel = 1
@@ -71,7 +69,8 @@ type InitOpts struct {
 	ComputeProviderID int
 	Throttle          bool
 	Scrypt            ScryptParams
-	ComputeBatchSize  uint64
+	// ComputeBatchSize must be greater than 0
+	ComputeBatchSize uint64
 }
 
 type ScryptParams struct {
@@ -136,8 +135,8 @@ func Validate(cfg Config, opts InitOpts) error {
 		return fmt.Errorf("invalid `opts.MaxFileSize`; expected: >= %d, given: %d", minFileSize, opts.MaxFileSize)
 	}
 
-	if opts.ComputeBatchSize == 0 || opts.ComputeBatchSize%8 != 0 {
-		return fmt.Errorf("invalid `opts.ComputeBatchSize` expected: > 0 and divisible by 8, given: %d", opts.ComputeBatchSize)
+	if opts.ComputeBatchSize == 0 {
+		return fmt.Errorf("invalid `opts.ComputeBatchSize` expected: > 0, given: %d", opts.ComputeBatchSize)
 	}
 
 	if res := shared.Uint64MulOverflow(cfg.LabelsPerUnit, uint64(opts.NumUnits)); res {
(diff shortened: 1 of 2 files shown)


dshulyak (approved):

config/config.go
```diff
@@ -136,6 +136,10 @@ func Validate(cfg Config, opts InitOpts) error {
 		return fmt.Errorf("invalid `opts.MaxFileSize`; expected: >= %d, given: %d", minFileSize, opts.MaxFileSize)
 	}
 
+	if opts.ComputeBatchSize == 0 || opts.ComputeBatchSize%8 != 0 {
+		return fmt.Errorf("invalid `opts.ComputeBatchSize` expected: > 0 and divisible by 8, given: %d", opts.ComputeBatchSize)
```
poszu: Why must a batch size be a multiple of 8?

NIT:
```suggestion
		return fmt.Errorf("invalid `opts.ComputeBatchSize` expected: > 0 and multiple of 8, given: %d", opts.ComputeBatchSize)
```

initialization/initialization_test.go
```diff
@@ -834,6 +834,45 @@ func TestStop(t *testing.T) {
...
+	opts.ComputeBatchSize = 0
+
+	_, err := NewInitializer(
+		WithNodeId(nodeId),
+		WithCommitmentAtxId(commitmentAtxId),
+		WithConfig(cfg),
+		WithInitOpts(opts),
+		WithLogger(testLogger{t: t}),
+	)
+	assert.Error(t, err)
```
poszu: It's good to verify that the expected error happened. You can do it with `assert.ErrorContains`

config/config.go
piersy: Hi @poszu that is what the doc says for that parameter.

https://github.com/spacemeshos/post/blob/eee6c3c1f18ce80f80de1a4c9fea834dfbade8a2/config/config.go#L15

Also I'm not understanding your nit suggestion, the two messages seem equivalent.

initialization/initialization_test.go
piersy: Hi @posu. I actually intentionally didn't check the content of the error.

My reasoning is as follows:

The content of the error is not part of the public api, so testing for it will make the test more brittle than it needs to be. From the perspective of the api what is important is that if an invalid compute batch size is provided,  an error is returned. Contrast that with something like [io.Reader](https://pkg.go.dev/io#Reader) where the EOF error is part of the api and callers actually make decisions based on the specific error returned.

What do you think? Happy to change it if you disagree?

config/config.go
poszu: I think that comment is outdated. In the past it was assumed that a label can be 1-256 bits wide. Now it's fixed to 16 bytes. The output data will always be a multiple of a byte.

Regarding the nit. For me, it's easier to think in terms of something being a multiple of 8, that's it.

initialization/initialization_test.go
poszu: I agree that in general, it's bad practice to look at the "text representation" of error in tests as it is subject to change and becomes irritating during refactors. However, without checking what kind of error was returned you can't really tell if the test works as intended - it might return a different error because some other input data is wrong - how can you tell?

piersy: It's not necessary to know (it's not necessary to ascertain which code path of the implementation is being executed), as long as for all invalid inputs an error is returned and for valid inputs no error is returned, then the component is functioning correctly.

So for testing a validation function I would first check that given valid input no error is returned, and then take the valid input and invalidate one field and call the validation function again to see that an error is returned and repeat that for all fields. Then I know that the validation function is working as expected, without having to put any further restrictions on its implementation.

It's true that maybe the error is wrong (has the wrong wording or is referencing some other problem) but similarly to how people don't generally test logging output, I don't think its particularly helpful to check errors at that level of detail.

config/config.go
piersy: Ok, so as long as compute batch size is greater than zero then it's valid?

poszu: I think it's fine.

poszu (approved):

PR #129, opened 2023-04-24
Bump post-rs to v0.1.6

poszu:


proving/proving.go
```diff
@@ -22,7 +22,7 @@ const (
 func Generate(ctx context.Context, ch shared.Challenge, cfg config.Config, logger shared.Logger, opts ...OptionFunc) (*shared.Proof, *shared.ProofMetadata, error) {
 	options := option{
 		threads:   1,
-		nonces:    20,
```
piersy, line 25: Is this change meant to be part of this PR, if so why?

poszu: Yes, because the proving algorithm changed, and the default parallel nonces changed from 20 to 16.

piersy (approved):
Looks good to me!

PR #131, opened 2023-04-25
OpenCL initialization integration

### LINKED ISSUE #130: Integration of OpenCL initialiser into `post` and `go-spacemesh`
### Description

The goal is to replace the existing `gpu-post` integration with the newly written OpenCL intitializer

### Acceptance Criteria

- Existing `gpu-post` integration is disabled, not removed (in case it is still needed). Switching between OpenCL and `gpu-post` is OK to require a rebuild of `post`.
- Test if go-spacemesh can initialize using OpenCL instead of `gpu-post`

fasmat:
Part of #130 

This integrates the OpenCL backend for initalization into the post repository. `gpu-post` has not been removed, but code using it has been altered to use the new `post-rs` API.

⚠️ Before merging this PR: merge https://github.com/spacemeshos/post-rs/pull/35 and release a new version of `post-rs` including the changes.


internal/postrs/bridge.go
```diff
@@ -0,0 +1,270 @@
...
+	defer C.free(cOut)
+
+	var cIdxSolution C.uint64_t
+	retVal := C.initialize(init, cStartPosition, cEndPosition, (*C.uint8_t)(cOut), &cIdxSolution)
+	if err := InitResultToError(retVal); err != nil {
+		return nil, nil, err
+	}
+
+	var vrfNonce *uint64
+	if cIdxSolution != 0 { // TODO(mafa): since 0 could be a valid nonce, we should find a better way to indicate no solution (e.g. InitializeOk = no solution, InitializeOkPow = solution)
```
fasmat, line 115: It appears that `C.initialize` returns -1 for `cIdxSolution` if none was found, double check.

internal/postrs/bridge.go
```diff
@@ -0,0 +1,266 @@
...
+	case C.InitializeFailedToGetProviders:
+		return ErrFetchProviders
+	default:
+		return fmt.Errorf("unknown error")
+	}
+}
+
+func cScryptPositions(opt *option) ([]byte, *uint64, error) {
+	gpuMtx.Lock()
+	defer gpuMtx.Unlock()
```
fasmat, line 83: I would like to use a `TryLock()` here to return early with an error ("device busy"), but this breaks a whole lot of tests in `go-spacemesh` that are running in parallel (and probably shouldn't).

internal/postrs/bridge.go
```diff
@@ -0,0 +1,266 @@
...
+		*vrfNonce = uint64(cIdxSolution)
+	}
+
+	output := C.GoBytes(cOut, C.int(cOutputSize))
+	return output, vrfNonce, nil
+}
+
+func cGetProviders() ([]ComputeProvider, error) {
+	gpuMtx.Lock()
+	defer gpuMtx.Unlock()
```
fasmat, line 123: Same here as above

internal/postrs/log.go
```diff
@@ -0,0 +1,33 @@
...
+import "C"
+
+import (
+	"github.com/spacemeshos/post/shared"
+)
+
+var log shared.Logger
+
+//export logCallback
+type logCallback func(level C.int, message *C.char)
```
fasmat: This is just a proof of work for a sugared logger. A real structured logger is a bit more complex, we would require more than just one callback function. A simple `C` structured logger we can take inspirations from is https://github.com/briandowns/struCtured-logger

@poszu do you want to log data from Rust or just messages? If it's just messages a simple interface like this is probably sufficient. If you want to log real data, that gets serialized to a proper json we should sync on how the interface should look like 🙂

Makefile.Inc
```diff
@@ -79,35 +78,30 @@ $(PROJ_DIR)$(GPU_SETUP_ZIP):
 get-gpu-setup: $(PROJ_DIR)$(GPU_SETUP_ZIP) $(BINDIR_GPU_SETUP_LIBS)
 .PHONY: get-gpu-setup
 
-
-####################################################
-# post-rs (https://github.com/spacemeshos/post-rs) #
-####################################################
+POSTRS_SETUP_REV = 0.1.7-beta3
```
poszu:
```suggestion
POSTRS_SETUP_REV = 0.1.8
```

internal/postrs/api.go
```diff
@@ -0,0 +1,122 @@
...
+func WithScryptN(n uint32) OptionFunc {
+	return func(opts *option) error {
+		opts.n = n
+		return nil
+	}
+}
+
+func WithVRFDifficulty(difficulty []byte) OptionFunc {
+	return func(opts *option) error {
+		if len(difficulty) != 32 {
```
poszu, line 91: The check wouldn't be necessary if the segment was [32]byte.

poszu (changes requested):
Partial review. Will continue later today.

Requesting changes for post-rs version in makefile.

internal/postrs/api.go
fasmat: Yes, but then the caller would need to do the check instead, unless we can somehow ensure that difficulty is `[32]byte` everywhere. Generally I prefer `[]byte` over `[32]byte` because converting from `[32]byte` to `[]byte` is cheap:

```golang
var param [32]byte
WithVRFDifficulty(param[:]) // [:] does not copy the underlying array
```

but the other way round is more involved and expensive:

```golang
param := make([]byte, 32)
var difficulty [32]byte
if len(param) != 32 {
    // in case len(param) is not guaranteed to be 32
    return fmt.Errorf("len(param) not 32")
}

copy(difficulty[:], param) // copy necessary
WithVrfDifficulty(difficulty)
```

The function calculating the difficulty also returns `[]byte` and changing it to return `[32]byte` just shifts the converting code into it.

poszu: What other difficulties do you have in mind? If there is another kind of difficulty somewhere else, then it might have a different length or even type. I don't see problem here.

IMO an API accepting an array as argument is self descriptive and clear - user doesn't need to figure out what the difficulty is and it doesn't need to be documented. The runtime check is moved to compile time. An invalid code won't compile instead of failing in runtime.

Regarding conversion slice ->array, user can just use an array and avoid coversion.

fasmat: There are no other difficulties, the difficulty is calculated for the size of the PoST during runtime here:

https://github.com/spacemeshos/post/blob/9393f007d9475c860a8cedf13d20ea4fd41caad5/shared/shared.go#L51-L58

`(*big.Int).FillBytes` takes a `[]byte` as parameter, we would need to then copy the result into a new variable of type `[32]bytes` and return this. `[]byte` are passed via reference, `[32]bytes` are passed via value unless we use `*[32]byte`. 

The check here is just a safety precaution (and cheap -> `len([]byte)` is `O(1)`). In our code it will never be a different length, but to enforce that during compile time makes the code less efficient, harder to read and less convenient to use imo.

fasmat: See also https://go.dev/doc/effective_go#arrays and https://go.dev/blog/slices-intro

poszu: I'm almost sure it's optimized nicely. Nevertheless, who cares about efficiency here? It's not a hot path.

IMHO, the code using array is actually more readable because the constraints are clearly presented in the API, which is more important. In other words, the API is easier to use.

initialization/initialization.go
```diff
@@ -457,27 +458,29 @@ func (init *Initializer) initFile(ctx context.Context, fileIndex int, batchSize,
...
-
-			init.nonce.Store(res.Nonce)
-			init.saveMetadata()
+			nonceValue := make([]byte, 16)
+			copy(nonceValue, res.Output[(*res.Nonce-startPosition)*16:])
+			init.logger.Info("initialization: file #%v, found nonce: %d, value: %x", fileIndex, *res.Nonce, nonceValue)
+
+			if init.nonceValue == nil || bytes.Compare(nonceValue, init.nonceValue) < 0 {
+				init.logger.Info("initialization: file #%v, found new best nonce", fileIndex)
+				init.nonceValue = nonceValue
```
poszu: NIT, the copy could be moved into the `if` to avoid alloc & copying if the condition is not met.

internal/postrs/api_test.go
```diff
@@ -0,0 +1,179 @@
...
+func TestScryptPositions(t *testing.T) {
+	providers, err := OpenCLProviders()
+	require.NoError(t, err)
+
+	vrfDifficulty := make([]byte, 32)
+	copy(vrfDifficulty, defaultDifficulty)
+	vrfDifficulty[0] = 0
+
+	start := uint64(1)
+	end := uint64(1 << 8)
```
poszu, line 47: I see this pattern in many places, I'm curious - why present some numbers as `1<< n`? For me, it results in more cognitive load compared to simply `end := 256`. Besides, there is no requirement for the number of labels to be a power of two.

internal/postrs/initializer.go
```diff
@@ -0,0 +1,155 @@
...
+		return ErrInvalidArgument
+	case C.InitializeFailedToGetProviders:
+		return ErrFetchProviders
+	default:
+		return ErrUnknown
+	}
+}
+
+// cScryptPositions calls the C functions from libpostrs that create the labels
+// and VRF proofs. It returns them as well as well as a possible error.
```
poszu:
```suggestion
// and VRF proofs. It returns them as well as a possible error.
```

internal/postrs/log.go
poszu: I was thinking about a simpler approach for now. Doing something more sophisticated would be overkill IMHO. Perhaps in the future? I was thinking about something like this: the callback would get a pointer to a structure containing some metadata (level, file, line, etc., and message string (already formatted)).

```go
// actually defined in C headers
type LogRecord struct {
  Level int
  Line int
  File String
  Message String
}

type logCallback func(record *LogRecord)
```

poszu: The other side is not included in post-rs yet and the API is not defined yet. Perhaps restrain from including it already until post-rs is ready?

internal/postrs/proof.go
```diff
@@ -30,9 +30,12 @@ func translateScryptParams(params config.ScryptParams) C.ScryptParams {
 	}
 }
 
-func GenerateProof(datadir string, challenge []byte, cfg config.Config, nonces uint, threads uint, powScrypt config.ScryptParams) (*shared.Proof, error) {
-	datadirPtr := C.CString(datadir)
-	defer C.free(unsafe.Pointer(datadirPtr))
+func GenerateProof(dataDir string, challenge []byte, cfg config.Config, nonces uint, threads uint, powScrypt config.ScryptParams) (*shared.Proof, error) {
+	// TODO(mafa): disabled for now (calling it more than once crashes the program)
+	// C.configure_logging(C.Trace)
```
poszu: The idea was that `C.configure_logging()` should be called only once, on "start"

poszu (approved):

internal/postrs/api.go
fasmat: I'm not agreeing with it makes the API easier to use, it might be idiomatic in Rust but in Go arrays are rarely used and imo this is not one of the places where it makes sense to use it.

Having the `WithVRFDifficulty` take a `[32]byte` instead of a `[]byte` requires the following changes:

Change this function https://github.com/spacemeshos/post/blob/9393f007d9475c860a8cedf13d20ea4fd41caad5/shared/shared.go#L51-L58

to 

```golang
func PowDifficulty(numLabels uint64) [32]byte { 
 	x := new(big.Int).Lsh(big.NewInt(1), 256) 
 	x.Div(x, big.NewInt(int64(numLabels))) 
 	x.Lsh(x, 3) // reduce difficulty by a factor of 8 
  
 	difficulty := make([]byte, 32)
 	x.FillBytes(difficulty)
	var result [32]byte
	copy(result[:], difficulty)
	return result
 }
```

and then where it is actually used:

https://github.com/spacemeshos/post/blob/9393f007d9475c860a8cedf13d20ea4fd41caad5/internal/postrs/initializer.go#L95-L100

convert it back to a slice?

```golang
	difficulty := opt.vrfDifficulty[:]
	cDifficulty := C.CBytes(difficulty)
	defer C.free(cDifficulty)
	init := C.new_initializer(cProviderId, cN, (*C.uchar)(cCommitment), (*C.uchar)(cDifficulty))
	if init == nil {
		return nil, nil, ErrInvalidProviderID
	}
```

> Nevertheless, who cares about efficiency here? It's not a hot path.

To some degree it is. To verify VRF nonces of incoming ATXs I have to initialize the one label that is referenced by the nonce and pass in the required difficulty (see `verifying.VerifyVRFNonce`). Depending on how many smeshers are participating in the network this might be called quite often - also consider that if a smesher wants to initialize a large PoST they probably have to do this over multiple epochs and keep updating their nonces 🙂

internal/postrs/api_test.go
fasmat: It carried over from the previous implementation and I personally have no issues with thinking in powers of two. I assume it might have been some performance thing where `gpu-post` was initializing in batches that were a power of 2 in size?

I don't have strong feelings either way, I just find with larger values, numbers of the kind `uint64(1 << 32)` become easier to parse visually and compare to each other than numbers of the kind `uint64(4000000000)`.

Do you want me to change the tests?

internal/postrs/log.go
fasmat: That is fine by me too, depending on what you mean with formatted string. With structured logging I meant giving `postrs` the ability to log data not just a position and a string:

```
{
    "level": "debug",
    "file": "init.rs",
    "line": 1234,
    "msg": "found new vrf nonce",
    "commitment": "483d1413ae0cafd2b881fbedff00ac8ecb9e58dc193b7a2daef135c9f620380f",
    "nonce": 9876,
    "label_value": "00009d3292ae3d6097747f5d220d736da0fa30844e3f931e22f56747b10305c8"
}
```

vs.

```
{
    "level": "debug",
    "file": "init.rs",
    "line": 1234,
    "msg": "found new vrf nonce; commitment: 483d1413ae0cafd2b881fbedff00ac8ecb9e58dc193b7a2daef135c9f620380f, nonce: 9876"
}
```

internal/postrs/proof.go
fasmat: I assumed so, I want to avoid making `GenerateProof`, `VerifyProof` and `Initialize` stateful, so I like the callback approach better 🙂

fasmat: Now that I look at it, this function can take a quite significant time before it returns. Shall we consider making it cancelable in some way? Verifying a proof is fast and initialization can be aborted between batches.

I know resuming an aborted `GenerateProof` is not possible - at least not without bigger changes - but now if a node is generating a proof I believe it cannot be shut down gracefully. At least not before it finishes proof generation.

PR #137, opened 2023-05-11
Log callback functions

fasmat:
This integrates the changes from https://github.com/spacemeshos/post-rs/pull/47/ to allow `postrs` to log while being used in `go-spacemesh`.


internal/postrs/proof.go
```diff
@@ -30,7 +32,11 @@ func translateScryptParams(params config.ScryptParams) C.ScryptParams {
 	}
 }
 
-func GenerateProof(dataDir string, challenge []byte, cfg config.Config, nonces uint, threads uint, powScrypt config.ScryptParams) (*shared.Proof, error) {
+func GenerateProof(dataDir string, challenge []byte, cfg config.Config, logger *zap.Logger, nonces uint, threads uint, powScrypt config.ScryptParams) (*shared.Proof, error) {
+	if logger != nil {
+		setLogCallback(logger)
+	}
```
poszu, line 38: I don't think that `GenerateProof()` and others should accept `logger` to be set for logging in the post-rs library. Only the first one will be set and remembered forever. While it's probably not a big deal in application code (as there will only be a single logger forever), in tests it will cause all logs to be recorded in the logger created for the first test that was executed probably causing them to appear in the wrong place (?).

I was imagining that there could be an exported function `func ConfigureLogging(logger *zap.Logger)` documented to be idempotent, which should be called once by the application on its boot (starting phase of postcli or go-spacemesh).

poszu (approved):

internal/postrs/proof.go
fasmat: I think a global logger in general is an issue. A `ConfigureLogging` function is a workaround for the main application, but our approach to logging is not to have one global instance but (possible multiple) instances that are dependency-injected into the services.

With a global logger running tests in parallel becomes more difficult because logs from different tests are all printed to the same output concurrently and cannot easily be associated with the test they came from. This is why for instance `zaptest.NewLog` takes an instance of `testing.T`, so that the logs can be sorted to the tests they came from and `go test` can print the outputs correctly.

EDIT: and tests can have different logging levels (e.g. tests where logs are irrelevant have a no-op logger, some have only info for dependent services and debug for the service at test, etc.)

Issue #133, opened 2023-04-30
post k2pow: change hashing algo and N


countvonzero:
## Description

quoting Iddo:

```
the scrypt+blake3 hash for k2pow will require that either all miners use GPU 
for proof generation, or lousy k1,k2 params and large ATX size for no good reason. 
The scrypt+sha3 that we use for labels is the worst thing that we can do because 
it has two extra disadvantages (efficient adversarial method for label computation 
will also break k2pow for free, and sha3 is inferior in general and we shouldn't use 
it in software ever).

We should just link to standard randomx repo (both rust crate and golang package 
are available),  it's easy to do in 5 minutes because it's totally unrelated to the GPU 
code, it's probably even easier than having two separate scrypt+sha3 functions 
and definitely easier than scrypt+blake3.
```

pigmej:
`use K2=37, K1=26`

countvonzero:
> `use K2=37, K1=26`

this change should be independent of k2/k1 number

EDIT: this K2=37/K1=26 should be for the scrypt+sha3 of k2pow hashing.

countvonzero:
from Iddo

```
the scryptN128+sha3 that we're using now is the worst possible hash, 
the k1,k2 params that we'd need with this scryptN128+sha3 are much 
lousier than what Tal gave (so our verification speed and ATX size will 
blow up to crazy complexities). The randomx hash is pretty much the 
best that we'd want (it's probably possible to do a littler better and indeed 
there are few existing alternatives, but randomx is standard and tested 
and very easy for us to integrate). Also randomx has very long history 
and it stood the test of time, there's nothing about say scrypt that's better 
in this regard, and say blake3 has an even shorter history. 

Again, the simple thing is that every agrees that randomx is strictly superior 
as the k2pow hash compared to scrypt.
```

countvonzero:
from Iddo

```
@pigmej @poszu  
 regarding the k2_difficulty param with randomx, on average modern CPU 
(such as my laptop ryzen 5850u) it's about 520 hashes/second with single core.
So I wrote an example code in rust with k2_difficulty that gives approximately 
the 2.5 minutes solving time that Tal wanted, here it is:
https://drive.google.com/file/d/1SwYQpofYf_Rx6l8yeZL01sf8clwP_6l-/view?usp=sharing
Ad here is the log when running it:
https://drive.google.com/file/d/1TCM7z9xHHRvsPSuJEPim52Zp1d-enE08/view?usp=sharing

[...]

PR #145, opened 2023-05-17
Remove K3 proof of work

poszu:
Integrates https://github.com/spacemeshos/post-rs/pull/68

Merge after 
- [x] https://github.com/spacemeshos/post-rs/pull/68 has been released and
- [x] https://github.com/spacemeshos/post/pull/137 has been merged

### DIFF
--- a/Makefile.Inc
+++ b/Makefile.Inc
@@ -49,7 +49,7 @@ else
 	endif
 endif
 
-POSTRS_SETUP_REV = 0.1.12-beta1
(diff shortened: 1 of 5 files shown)


countvonzero (approved):

fasmat:
Why is K3 Pow removed but not K3? If K3 == K2 why do we have it at all?

fasmat:
> Why is K3 Pow removed but not K3? If K3 == K2 why do we have it at all?

Answer can be found in this discussion https://github.com/spacemeshos/post/issues/133

PR #154, opened 2023-06-02
Integrate RandomX pow

poszu:
Integration of [RandomX](https://github.com/tevador/RandomX)-based proof of work in place of the scrypt-based one.

### Changes
- PoW difficulty is now 256bit, configured in form of `[32]byte` array;
- verifier is now stateful because RandomX cache takes about 250ms to initialize, which is too long to do it per ATX;
- proving and verifying options take `PowFlags` that configure the operation of RandomX. They have sane defaults:
  - proving runs in FAST mode (2GiB RAM)
  - verification runs in LIGHT mode (256MiB RAM)
- renamed "K2 PoW" to just "PoW" everywhere. Reasoning - there is only one PoW in POST now AND the old name was uninformative anyway.

:bulb: configuration of RandomX flags affect speed and memory requirements only - misconfiguring it won't affect the generated pow nonce correctness.

:bulb: `postrs.Verifier` can be used from many threads safely.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -79,17 +79,20 @@ type Config struct {
 	K2 uint32 // K2 is the number of labels below the required difficulty required for a proof.
 	K3 uint32 // K3 is the size of the subset of proof indices that is validated.
 
-	PowDifficulty [32]byte
+	// FIXME: remove support for the old scrypt-based PoW
+	K2PowDifficulty uint64
+	PowDifficulty   [32]byte
 }
 
 func DefaultConfig() Config {
 	cfg := Config{
-		LabelsPerUnit: 512, // 8kB units
-		MaxNumUnits:   defaultMaxNumUnits,
-		MinNumUnits:   defaultMinNumUnits,
-		K1:            26,
-		K2:            37,
-		K3:            37,
+		LabelsPerUnit:   512, // 8kB units
+		MaxNumUnits:     defaultMaxNumUnits,
+		MinNumUnits:     defaultMinNumUnits,
+		K1:              26,
+		K2:              37,
+		K3:              37,
+		K2PowDifficulty: 0x0FFFFFFF_FFFFFFFF,
 	}
 	for i := range cfg.PowDifficulty {
 		cfg.PowDifficulty[i] = 0xFF
@@ -124,6 +127,14 @@ func (p *ScryptParams) Validate() error {
 		return errors.New("scrypt parameter p cannot be 0")
 	}
 	return nil
+}
+
+func DefaultPowParams() ScryptParams {
+	return ScryptParams{
+		N: 128,
+		R: 1,
+		P: 1,
+	}
 }
 
 func DefaultLabelParams() ScryptParams {
--- a/config/config.go
+++ b/config/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/spacemeshos/post/internal/postrs"
 	"github.com/spacemeshos/post/shared"
 )
 
@@ -36,6 +37,39 @@ func init() {
 	DefaultDataDir = filepath.Join(home, "post", DefaultDataDirName)
 }
 
+type PowFlags = postrs.PowFlags
+
+const (
+	// Use the full dataset. AKA "Fast mode".
+	PowFastMode = postrs.PowFastMode
+	// Allocate memory in large pages.
+	PowLargePages = postrs.PowLargePages
+	// Use JIT compilation support.
+	PowJIT = postrs.PowJIT
+	// When combined with FLAG_JIT, the JIT pages are never writable and executable at the same time.
+	PowSecure = postrs.PowSecure
+	// Use hardware accelerated AES.
+	PowHardAES = postrs.PowHardAES
+	// Optimize Argon2 for CPUs with the SSSE3 instruction set.
+	PowArgon2SSSE3 = postrs.PowArgon2SSSE3
+	// Optimize Argon2 for CPUs with the SSSE3 instruction set.
+	PowArgon2AVX2 = postrs.PowArgon2AVX2
+	// Optimize Argon2 for CPUs without the AVX2 or SSSE3 instruction sets.
+	PowArgon2 = postrs.PowArgon2
+)
+
+func RecommendedPowFlags() PowFlags {
+	return postrs.GetRecommendedPowFlags()
+}
+
+func DefaultProvingPowFlags() PowFlags {
+	return RecommendedPowFlags() | PowFastMode
+}
+
+func DefaultVerifyingPowFlags() PowFlags {
+	return RecommendedPowFlags()
+}
+
 type Config struct {
 	MinNumUnits   uint32
 	MaxNumUnits   uint32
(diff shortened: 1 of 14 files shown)


internal/postrs/proof.go
```diff
@@ -44,10 +46,11 @@ func GenerateProof(dataDir string, challenge []byte, cfg config.Config, logger *
...
-		k1:                C.uint32_t(cfg.K1),
-		k2:                C.uint32_t(cfg.K2),
-		k2_pow_difficulty: C.uint64_t(cfg.K2PowDifficulty),
-		pow_scrypt:        translateScryptParams(powScrypt),
+		k1: C.uint32_t(K1),
+		k2: C.uint32_t(K2),
+	}
+	for i, b := range powDifficulty {
+		config.pow_difficulty[i] = C.uchar(b)
 	}
```
fasmat, line 54: Why not do the same for the difficulty as for the challenge?

```golang
	diffPtr := C.CBytes(powDifficulty)
	defer C.free(diffPtr)

	config := C.Config{
		k1: C.uint32_t(K1),
		k2: C.uint32_t(K2),
		pow_difficulty: (*C.uchar)(diffPtr),
	}
```

avoids copying the data, but requires `pow_difficulty` to be a pointer instead of an array.

proving/proving.go
```diff
@@ -16,9 +16,9 @@ import (
 
 func Generate(ctx context.Context, ch shared.Challenge, cfg config.Config, logger *zap.Logger, opts ...OptionFunc) (*shared.Proof, *shared.ProofMetadata, error) {
 	options := option{
-		threads:   1,
-		nonces:    16,
-		powScrypt: config.DefaultPowParams(),
+		threads:  1,
+		nonces:   16,
+		powFlags: postrs.GetRecommendedPowFlags() | postrs.PowFastMode,
```
fasmat:
```suggestion
		powFlags: config.DefaultProvingPowFlags(),
```

config/config.go
```diff
@@ -45,19 +79,26 @@ type Config struct {
...
+		K1:              26,
+		K2:              37,
+		K3:              37,
 		K2PowDifficulty: 0x0FFFFFFF_FFFFFFFF,
 	}
+	for i := range cfg.PowDifficulty {
+		cfg.PowDifficulty[i] = 0xFF
+	}
+
+	return cfg
```
fasmat, line 101: With `PowDifficulty` as a slice this can be written more concisely as:

```golang
	return Config{
		LabelsPerUnit:   512, // 8kB units
		MaxNumUnits:     defaultMaxNumUnits,
		MinNumUnits:     defaultMinNumUnits,
		K1:              26,
		K2:              37,
		K3:              37,
		K2PowDifficulty: 0x0FFFFFFF_FFFFFFFF,
		PowDifficulty:   bytes.Repeat(0xFF, 32),
	}
}
```

The check for the length then has to be done in `GenerateProof` and `VerifyProof` (where we are already checking for other lengths as well).

It would also allow to pass the difficulty without the need to copy it first (see other comment).

internal/postrs/proof.go
```diff
@@ -69,11 +73,78 @@ func GenerateProof(dataDir string, challenge []byte, cfg config.Config, logger *
...
+	if result != C.Ok {
+		return nil, fmt.Errorf("failed to create verifier")
+	}
+	return &verifier, nil
+}
+
+func (v *Verifier) Close() error {
+	v.closeOnce.Do(func() { C.free_verifier(v.inner) })
+	return nil
+}
```
fasmat, line 132: Should multiple calls to close return an error?

verifying/verifying.go
```diff
@@ -59,9 +59,35 @@ func VerifyVRFNonce(nonce *uint64, m *shared.VRFNonceMetadata, opts ...OptionFun
 	return nil
 }
 
+type ProofVerifier struct {
+	inner *postrs.Verifier
+}
```
fasmat:
```suggestion
type ProofVerifier struct {
	*postrs.Verifier
}
```

With struct embedding `ProofVerifier` inherits the methods of `postrs.Verifier` (no need to define a `Close` method that just calls the inner `Close`)

verifying/verifying.go
```diff
@@ -59,9 +59,35 @@ func VerifyVRFNonce(nonce *uint64, m *shared.VRFNonceMetadata, opts ...OptionFun
...
+
+// NewProofVerifier creates a new proof verifier.
+// If powFlags is nil, the recommended PoW flags will be used.
+// The verifier must be closed after use with Close().
+func NewProofVerifier(powFlags *config.PowFlags) (*ProofVerifier, error) {
```
fasmat: I think we should keep the functional options for optional parameters and not use `nil` as a placeholder for "default parameter".

We are already using them in the `proving` package and adding them here in the `verifying` package would make a more consistent API.

internal/postrs/proof.go
poszu: No, I wanted it to be idempotent

internal/postrs/proof.go
poszu: It doesn't avoid copying the data because `C.CBytes(powDifficulty)` allocates memory and makes a copy. It will only add a layer of indirection and lose the benefit of having the length embedded in the type (`[32]byte` vs `[]byte`), which makes the API easier to use properly.

config/config.go
poszu: I prefer to put the required length into the type. Then, the length is obvious and manual validation is not required.

internal/postrs/proof.go
fasmat: `[32]byte` is a value type so it is copied when `C.generate_proof` function is called a bit further down, while `C.CBytes` returns a pointer to the data that doesn't need to be copied again.

I don't think it makes much difference between arrays and slices in terms of what is easier to use. We even use `[]byte` over `[x]byte` for most other types here, passing fewer bytes to the function also just fils them with zeros instead of causing an error during compilation: https://go.dev/play/p/i_5jDu8yc50

poszu: >[32]byte is a value type so it is copied when C.generate_proof function is called a bit further down, while C.CBytes returns a pointer to the data that doesn't need to be copied again.

That's the same amount of copying, except `C.CBytes()` allocates on a heap (and copies the contents into there) while the array is kept on a stack. Anyhow, it's supposed to be called once every epoch.

>I don't think it makes much difference between arrays and slices in terms of what is easier to use.

I mean that an array, which is fixed length better expresses the API requirements.
- `[32]byte` says: "Pass me **exactly 32** bytes." 
- `[]byte` says: "Pass me **some** bytes

In other words, when I see a function signature:
```go
func Foo(val [32]byte)
```
- I immediately know that I must pass 32 bytes,
- I don't need to validate length of passed data (which might be necessary if `[]byte` is passed).

fasmat: > That's the same amount of copying, except C.CBytes() allocates on a heap (and copies the contents into there) while the array is kept on a stack.

Array does 2 copies: here in this `for` loop and again when calling `C.generate_proof` by passing the array via value to the function.

Slice only once: when calling `C.CBytes`, after that only the pointer to the array is passed to `C.generate_proof`.

fasmat: But you are right it's not a necessary optimization because generating a proof is only done every 2 weeks and compared to the rest of the process its an negligible improvement.

fasmat (approved):

PR #165, opened 2023-07-03
Create a release with postcli when a tag is pushed

### LINKED ISSUE #124: Build and produce artifacts for postcli
There is a big demand for postcli. We should include postcli as a binary so users don't have to compile it by hand.

poszu:
Closes #124 

Example release: https://github.com/spacemeshos/post/releases/tag/untagged-c596b7c32e5a556bcc89
Example CI run: https://github.com/spacemeshos/post/actions/runs/5444871501

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,23 +141,10 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
       - name: Download artifacts
         uses: actions/download-artifact@v3
         with:
           path: ./artifacts
-      - name: List artifacts
-        run: ls -R ./artifacts
-      - name: Create a draft release
-        id: create_release
-        uses: actions/create-release@v1
-        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - develop
+    tags:
+      - 'v*'
```
fasmat, line 9: There is a dedicated trigger that runs jobs when a release is created on a GH repository:

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

Might be cleaner than triggering on tags.

fasmat (approved):

.github/workflows/ci.yml
poszu: It depends on what is more ergonomic, which depends on the person :shrug:. I personally like to just push a tag and collect the automatically created release a few minutes later, without clicking over the web GUI. Ideally, the workflow could support both.

PR #166, opened 2023-07-03
Run unit tests on arm64 runners

### LINKED ISSUE #161: Execute unit tests on arm64 runners
#### Acceptance criteria

- Unit test job of post as well as the release jobs are also executed on `linux/arm64` and `darwin/arm64`

#### Implementation hints

- Use self hosted VMs already available for the organization.

fasmat:
Closes #161 

I had to stop parallization of initialization test because it slows down the test significantly: Ubuntu 1m40s with `t.Parallel`, 1m20s without; Linux arm64 timeout with, 1m56s without.

Initialization in parallel isn't really supported anyway. Any call to `Initialize` will wait if another goroutine is already initializing with the selected provider.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: 
+          - ubuntu-latest
+          - [self-hosted, linux, arm64]
+          - macos-latest
+          - [self-hosted, macos, arm64]
+          - windows-latest
     steps:
       - name: checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -54,7 +54,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: 
+          - ubuntu-latest
+          - [self-hosted, linux, arm64]
+          - macos-latest
+          - [self-hosted, macos, arm64]
```
poszu, line 61: RandomX PoW (executed in `GenerateProof()`) might crash on mac arm64 without the SDK override.

poszu (approved):

PR #167, opened 2023-07-04
Increase build timeout

fasmat:
Last build failed because it timed out. Reason is on windows setting up the environment already takes almost 10 minutes so the build itself doesn't finish in time.


poszu (approved):
Perhaps the Windows Defender "fix" would help?

PR #169, opened 2023-07-04
Update CI tooling

fasmat:
Updates gotestsum, golangci-lint and staticcheck to their most recent versions.

Additionally the steps were reordered: the setup (installing dependencies, configuring the VM) is done before the checkout. This speeds up the CI especially on Windows where checking out the code without an active Windows Defender is about twice as fast as with it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,15 +61,6 @@ jobs:
           - [self-hosted, macos, arm64]
           - windows-latest
     steps:
-      - name: checkout
-        uses: actions/checkout@v3
-      - name: set up go
-        uses: actions/setup-go@v4
-        with:
-          check-latest: true
-          go-version-file: "go.mod"
-      - name: setup env
-        run: make install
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -83,6 +74,15 @@ jobs:
...
         run: choco install opencl-intel-cpu-runtime
+      - name: checkout
+        uses: actions/checkout@v3
+      - name: set up go
+        uses: actions/setup-go@v4
+        with:
+          check-latest: true
+          go-version-file: "go.mod"
+      - name: setup env
+        run: make install
```
countvonzero, line 85: what is the effect of changing the orders? these are also not described in the PR description.

fasmat: the idea is to prepare the VM/runner before checking out the code. Especially on windows this makes a big difference, first disabling the Windows defender and then checking out code is faster than the other way round (because Windows defender slows down the checkout).

Added to PR description.

countvonzero (approved):

PR #173, opened 2023-07-10
Optional miner ID for k2pow input

poszu:
Part of https://github.com/spacemeshos/go-spacemesh/issues/4656

### DIFF
--- a/Makefile.Inc
+++ b/Makefile.Inc
@@ -49,7 +49,7 @@ else
 	endif
 endif
 
-POSTRS_SETUP_REV = 0.4.0-rc2
+POSTRS_SETUP_REV = 0.4.0
 POSTRS_SETUP_ZIP = libpost-$(platform)-v$(POSTRS_SETUP_REV).zip
 POSTRS_SETUP_URL_ZIP ?= https://github.com/spacemeshos/post-rs/releases/download/v$(POSTRS_SETUP_REV)/$(POSTRS_SETUP_ZIP)
 ifeq ($(platform), windows)
--- a/Makefile.Inc
+++ b/Makefile.Inc
@@ -49,7 +49,7 @@ else
 	endif
 endif
 
-POSTRS_SETUP_REV = 0.3.1
+POSTRS_SETUP_REV = 0.4.0-rc2
 POSTRS_SETUP_ZIP = libpost-$(platform)-v$(POSTRS_SETUP_REV).zip
 POSTRS_SETUP_URL_ZIP ?= https://github.com/spacemeshos/post-rs/releases/download/v$(POSTRS_SETUP_REV)/$(POSTRS_SETUP_ZIP)
 ifeq ($(platform), windows)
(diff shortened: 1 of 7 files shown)


Makefile.Inc
```diff
@@ -49,7 +49,7 @@ else
 	endif
 endif
 
-POSTRS_SETUP_REV = 0.3.1
+POSTRS_SETUP_REV = 0.4.0-rc2
```
poszu: TODO: Update when post-rs is merged and released

internal/postrs/proof.go
```diff
@@ -34,7 +35,26 @@ func TranslateScryptParams(n, r, p uint) ScryptParams {
...
 }
 
-func GenerateProof(dataDir string, challenge []byte, logger *zap.Logger, nonces, threads uint, K1, K2 uint32, powDifficulty [32]byte, powFlags PowFlags) (*shared.Proof, error) {
+type postOptions struct {
+	powCreatorId []byte
+}
+
+type PostOptionFunc func(*postOptions) error
+
+func WithPowCreator(id []byte) PostOptionFunc {
```
fasmat, line 44: Depending on the effort, would it make sense to change all the other properties also to functional options?

internal/postrs/proof.go
```diff
@@ -138,7 +166,14 @@ func (v *Verifier) VerifyProof(
 	k1, k2, k3 uint32,
 	powDifficulty [32]byte,
 	scryptParams ScryptParams,
+	options ...PostOptionFunc,
```
fasmat, line 169: Same as for proof generation, would it make sense to change the other parameters too?

internal/postrs/proof.go
```diff
@@ -183,6 +218,16 @@ func (v *Verifier) VerifyProof(
 		},
 	}
 
+	if opts.powCreatorId != nil {
+		logger.With().Info("Verifying with PoW creator ID", zap.String("id", hex.EncodeToString(opts.powCreatorId)))
+		minerIdSliceHdr := (*reflect.SliceHeader)(unsafe.Pointer(&opts.powCreatorId))
```
fasmat, line 223: Why here not also just `C.Bytes(opts.PowCraetorId)` and free it on the go side afterwards?

`len` is always guaranteed to be 32 already checked before passed into this function and `cap` should be irrelevant for rust.

internal/postrs/proof.go
poszu: Yes, it makes sense. But it's an effort we don't have time for right now.

internal/postrs/proof.go
poszu: I followed the logic above already present for indices. It's nice as, in contrast to `C.Bytes()` it doesn't copy.

fasmat (approved):

internal/postrs/proof.go
fasmat: Created issue to address later: https://github.com/spacemeshos/post/issues/174

PR #177, opened 2023-07-12
Use mainnet params in postcli

fasmat:


pigmej (approved):
But pls cross check with @dshulyak

dshulyak (approved):

PR #179, opened 2023-07-12
Allow initializing only subset of labels

poszu:
## Motivation
It is now possible to initialize a subset of POS data with `postcli`. The initialization can be split on file boundary (i.e. files (0-N), (N+1 - M), etc.)

## Changes
Added 2 new arguments to `postcli`:
```
-fromFile int
        index of the first file to init (inclusive)
-toFile int
        index of the last file to init (inclusive). Will init to the end of declared space if not provided.
```

Added the value of found nonce to metadata in order to be able to select the best nonce. The operator combining separately initialized data will need to manually pick the best CRF nonce if they wish to have the best possible nonce. Details in the updated readme.

### DIFF
--- a/cmd/postcli/README.md
+++ b/cmd/postcli/README.md
@@ -48,6 +48,13 @@ https://github.com/intel/compute-runtime/releases
 ./postcli -printProviders
 ```
 
+###  Print the number of files that would be initialized
+
+```bash
+./postcli -numUnits 100 -printNumFiles
+1600
+```
+
 ### Print the used config and options
 
 ```bash
@@ -62,6 +69,59 @@ Example
 ./postcli -provider=2 -id=c230c51669d1fcd35860131e438e234726b2bd5f9adbbd91bd88a718e7e98ecb -commitmentAtxId=c230c51669d1fcd35860131e438e234726b2bd5f9adbbd91bd88a718e7e98ecb -genproof
 
 ```
+
+### Initialize subset
+It is possible to initialize only subset of the files. It's useful if one wants to split initialization between many machines.
+#### Example - split initialization between 2 machines
+Let's assume one wants to initialize 100 units and split the process of initialization to machine A and B.
+
+First let's see how many files it would be:
+```bash
+./postcli -numUnits 100 -printNumFiles
+1600
+```
+
+Next, on machine A:
+```bash
+./postcli -numUnits 100 -id <id> -commitmentAtxId <id> -toFile 799 -datadir ./dataA
+```
+We get the first half - 800 binary files
+```bash
+ls -la ./dataA/*.bin | wc -l
+800
+```
+
+Next, on machine B:
+```bash
+./postcli -numUnits 100 -id <id> -commitmentAtxId <id> -fromFile 800 -datadir ./dataB
+```
+We get the first half - 800 binary files
+```bash
+ls -la ./dataB/*.bin | wc -l
+800
+```
+
+Finally we can combine both sets together. A dummy example to get the feeling. Realisticly it would probably be copying over the network
+```bash
+cp ./dataA/* ./data/
+cp ./dataB/* ./data/
+ls -la ./data/*.bin | wc -l
+1600
+```
+
+**An optional step to select best possible VRF nonce**
+Normally, when `postcli`initializates from the start to the end it will automatically pick the best vrf nonce. The best means pointing to **the label with the smallest value**. It might be important in the future if we allow increasing POS data size, but it is not critical at the moment.
+
+Now, when `postcli` initializes in chunks, each subset will find a valid vrf nonce, which represents the local minimum in the inititalized subset. It's optimal to select the best one (the global minimum), but it's not critical.
+
+The values of nonces are 128bit, represented as a 16B binary array in big endian. Given two nonces:
+```
+nonceA = 0000ffda94993723a980bf557509773e
+nonceB = 0000488e171389cce69344d68b66f6b4
+```
+`nonceB` represents the smaller value.
+
+The nonce and the label it points to is included in the post_metadata.json. It's currently up to the operator to find the best VRF nonce manually and copy the `Nonce` and `NonceValue` to the postdata_metadata.json on the target machine.
 
 ### Remarks
 * `-id` and `-commitmentAtxId` are required because they are committed to the generated data.
--- a/cmd/postcli/README.md
+++ b/cmd/postcli/README.md
@@ -95,7 +95,7 @@ Next, on machine B:
 ```bash
 ./postcli -numUnits 100 -id <id> -commitmentAtxId <id> -fromFile 800 -datadir ./dataB
 ```
-We get the first half - 800 binary files
+We get the second half - 800 binary files
 ```bash
 ls -la ./dataB/*.bin | wc -l
 800
@@ -110,18 +110,20 @@ ls -la ./data/*.bin | wc -l
 ```
 
 **An optional step to select best possible VRF nonce**
-Normally, when `postcli`initializates from the start to the end it will automatically pick the best vrf nonce. The best means pointing to **the label with the smallest value**. It might be important in the future if we allow increasing POS data size, but it is not critical at the moment.
+Normally, when `postcli`initializates from the start to the end it will automatically pick the best VRF nonce. The best means pointing to **the label with the smallest value**. This will avoid a longer initialization time when a node increases their PoST size in the future (not supported yet).
 
-Now, when `postcli` initializes in chunks, each subset will find a valid vrf nonce, which represents the local minimum in the inititalized subset. It's optimal to select the best one (the global minimum), but it's not critical.
+Now, when `postcli` initializes in chunks, each subset will find a valid vrf nonce, which represents the local minimum in the inititalized subset. It is recommended to select the best one (the global minimum).
 
 The values of nonces are 128bit, represented as a 16B binary array in big endian. Given two nonces:
 ```
-nonceA = 0000ffda94993723a980bf557509773e
-nonceB = 0000488e171389cce69344d68b66f6b4
+NonceA = 12345
+NonceValueA = 0000ffda94993723a980bf557509773e
+NonceB = 98765
+NonceValueB = 0000488e171389cce69344d68b66f6b4
 ```
-`nonceB` represents the smaller value.
+`NonceB` is the global minimum since its value is smaller than the one of `NonceA`.
 
-The nonce and the label it points to is included in the post_metadata.json. It's currently up to the operator to find the best VRF nonce manually and copy the `Nonce` and `NonceValue` to the postdata_metadata.json on the target machine.
+The nonce (index) and noncevalue (the label) is included in the post_metadata.json. It is up to the operator to find the best VRF nonce manually and copy the `Nonce` and `NonceValue` to the postdata_metadata.json on the target machine.
 
 ### Remarks
 * `-id` and `-commitmentAtxId` are required because they are committed to the generated data.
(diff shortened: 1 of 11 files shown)


initialization/layout.go
```diff
@@ -1,17 +1,41 @@
...
+	From uint64
+	To   uint64
+
 	NumFiles          uint
 	FileNumLabels     uint64
 	LastFileNumLabels uint64
 }
 
-func deriveFilesLayout(cfg config.Config, opts config.InitOpts) filesLayout {
+func deriveFilesLayout(cfg config.Config, opts config.InitOpts) (*filesLayout, error) {
```
fasmat: filesLayout doesn't need to be a pointer, just return `filesLayout{}` in the error case.

shared/post_metadata.go
```diff
@@ -8,6 +13,31 @@ type PostMetadata struct {
...
+
+// MarshalJSON to hex
+func (n NonceValue) MarshalJSON() ([]byte, error) {
+	if n == nil {
+		return []byte{}, nil
+	}
+	dst := make([]byte, hex.EncodedLen(len(n)))
+	hex.Encode(dst, n[:])
+	return []byte(fmt.Sprintf("\"%s\"", dst)), nil
 }
```
fasmat: No need to make it more complicated then necessary 😉 

```suggestion
type NonceValue []byte

func (n *NonceValue) MarshalJSON() ([]byte, error) {
	return json.Marshal(hex.EncodeToString(*n))
}

func (n *NonveValue) UnmarshalJSON(data []byte) (err error) {
	var hexString string
	if err = json.Unmarshal(data, &hexString); err != nil {
		return
	}
	*n, err = hex.DecodeString(hexString)
	return
}
```

initialization/initialization.go
```diff
@@ -590,6 +603,7 @@ func (init *Initializer) saveMetadata() error {
 		NumUnits:        init.opts.NumUnits,
 		MaxFileSize:     init.opts.MaxFileSize,
 		Nonce:           init.nonce.Load(),
+		NonceValue:      init.nonceValue,
```
fasmat, line 604: Why is this value needed?

initialization/initialization.go
```diff
@@ -234,19 +234,32 @@ func (init *Initializer) Initialize(ctx context.Context) error {
...
+
 	init.logger.Info("initialization file layout",
 		zap.Uint("numFiles", layout.NumFiles),
 		zap.Uint64("labelsPerFile", layout.FileNumLabels),
 		zap.Uint64("labelsLastFile", layout.LastFileNumLabels),
+		zap.Int("firstFileIndex", firstFileIndex),
+		zap.Int("lastFileIndex", lastFileIndex),
 	)
-	if err := init.removeRedundantFiles(layout); err != nil {
+	if err := init.removeRedundantFiles(*layout); err != nil {
```
fasmat: no need to dereference if `deriveFilesLayout` returns a value 🙂

initialization/initialization.go
poszu: I explained in the description:
>Added the value of found nonce to metadata in order to be able to select the best nonce. A person/script joining separately initialized data will need to manually pick the best nonce (or not if they don't care).

poszu:
## An example of overlapping initializations and md5sums of the created data:
### ~/post/data4
Skip `-from` - start from the beginning.
```
go run ./cmd/postcli  -maxFileSize 1048576 -labelsPerUnit 262144 -to 262144 -provider 4294967295 -commitmentAtxId 00037ec8ec25e6d2c00000000000000000000000000000000000000000000000 -id 518e2824769a7e2ca844bdd6045109c875f62c7505944d79139a2c85aaa3de3c -datadir ~/post/data4/
❯ md5sum ~/post/data4/*
c812b7e913df797245fe6dc5b1cdede9  /home/bartosz/post/data4/postdata_0.bin
febee84428bb6bb3a61d44442a4198b1  /home/bartosz/post/data4/postdata_1.bin
66b66cd003a4bc3faf4721b8256e7f9c  /home/bartosz/post/data4/postdata_2.bin
22197e3c25e7491e714860e13a6fb1e4  /home/bartosz/post/data4/postdata_3.bin
```
### ~/post/data5
```
❯ go run ./cmd/postcli  -maxFileSize 1048576 -labelsPerUnit 262144 -from 65536 -to 327680 -provider 4294967295 -commitmentAtxId 00037ec8ec25e6d2c00000000000000000000000000000000000000000000000 -id 518e2824769a7e2ca844bdd6045109c875f62c7505944d79139a2c85aaa3de3c -datadir ~/post/data5/
❯ md5sum ~/post/data5/*
febee84428bb6bb3a61d44442a4198b1  /home/bartosz/post/data5/postdata_1.bin
66b66cd003a4bc3faf4721b8256e7f9c  /home/bartosz/post/data5/postdata_2.bin
22197e3c25e7491e714860e13a6fb1e4  /home/bartosz/post/data5/postdata_3.bin
af906c228c223ddecabf120ce76c7cc3  /home/bartosz/post/data5/postdata_4.bin
44dd75008fd9e6b7c99b50dd55e873ab  /home/bartosz/post/data5/postdata_metadata.jso
```

poszu:
I changed the interface. It now accepts `-fromFile <idx>` and `-toFile <idx>` - file indices to initialize. This is less error-prone and allows splitting init at a file boundary.

dshulyak:
does it allow to set file size?

pigmej:
@dshulyak `--max-file-size` ? :)

cmd/postcli/README.md
```diff
@@ -63,6 +70,59 @@ Example
...
+```bash
+ls -la ./dataA/*.bin | wc -l
+800
+```
+
+Next, on machine B:
+```bash
+./postcli -numUnits 100 -id <id> -commitmentAtxId <id> -fromFile 800 -datadir ./dataB
+```
+We get the first half - 800 binary files
```
fasmat:
```suggestion
We get the second half - 800 binary files
```

cmd/postcli/README.md
```diff
@@ -63,6 +70,59 @@ Example
...
+Finally we can combine both sets together. A dummy example to get the feeling. Realisticly it would probably be copying over the network
+```bash
+cp ./dataA/* ./data/
+cp ./dataB/* ./data/
+ls -la ./data/*.bin | wc -l
+1600
+```
+
+**An optional step to select best possible VRF nonce**
+Normally, when `postcli`initializates from the start to the end it will automatically pick the best vrf nonce. The best means pointing to **the label with the smallest value**. It might be important in the future if we allow increasing POS data size, but it is not critical at the moment.
```
fasmat:
```suggestion
Normally, when `postcli`initializates from the start to the end it will automatically pick the best VRF nonce. The best means pointing to **the label with the smallest value**. This will avoid a longer initialization time when a node increases their PoST size in the future (not supported yet).
```

cmd/postcli/README.md
```diff
@@ -63,6 +70,59 @@ Example
...
+
+Now, when `postcli` initializes in chunks, each subset will find a valid vrf nonce, which represents the local minimum in the inititalized subset. It's optimal to select the best one (the global minimum), but it's not critical.
```
fasmat:
```suggestion
Now, when `postcli` initializes in chunks, each subset will find a valid vrf nonce, which represents the local minimum in the inititalized subset. It is recommended to select the best one (the global minimum).
```

cmd/postcli/README.md
```diff
@@ -63,6 +70,59 @@ Example
...
+
+The values of nonces are 128bit, represented as a 16B binary array in big endian. Given two nonces:
+```
+nonceA = 0000ffda94993723a980bf557509773e
+nonceB = 0000488e171389cce69344d68b66f6b4
+```
+`nonceB` represents the smaller value.
```
fasmat:
````suggestion
```
NonceA = 12345
NonceValueA = 0000ffda94993723a980bf557509773e
NonceB = 98765
NonceValueB = 0000488e171389cce69344d68b66f6b4
```
`NonceB` is the global minimum since its value is smaller than the one of `NonceA`.
````

File: cmd/postcli/README.md [hunk 9]
```diff
@@ -63,6 +70,59 @@ Example
...
+
+The nonce and the label it points to is included in the post_metadata.json. It's currently up to the operator to find the best VRF nonce manually and copy the `Nonce` and `NonceValue` to the postdata_metadata.json on the target machine.
```
fasmat:
```suggestion
The nonce (index) and noncevalue (the label) is included in the post_metadata.json. It is up to the operator to find the best VRF nonce manually and copy the `Nonce` and `NonceValue` to the postdata_metadata.json on the target machine.
```

cmd/postcli/main.go
```diff
@@ -97,6 +109,12 @@ func main() {
 		return
 	}
 
+	if printNumFiles {
+		totalFiles := initialization.TotalFiles(cfg, opts)
```
fasmat: I think it makes sense to have totalfiles as method of `cfg` taking `opts` as parameter.

initialization/initialization_test.go
```diff
@@ -868,3 +849,116 @@ func initData(datadir string) ([]byte, error) {
...
+	r.NoError(err)
+	err = initSubset.Initialize(context.Background())
+	r.NoError(err)
+
+	// Verify that the subset is a subset of the full set
+	fullData, err := initData(opts.DataDir)
+	r.NoError(err)
+	subsetData, err := initData(optsSubset.DataDir)
+	r.NoError(err)
+	r.True(bytes.Contains(fullData, subsetData))
```
fasmat, line 898:
```suggestion
	r.Contains(fullData, subsetData)
```

poszu: It doesn't work unless I wrap both in `string()` :shrug:

fasmat: Maybe this instead?

```suggestion
	r.Equal(fullData[optsSubset.FromFileIdx*int(opts.MaxFileSize):(*optsSubset.ToFileIdx+1)*int(opts.MaxFileSize)], subsetData)
```

`byte.Contains` imo is a not strict enough check. if `subset` is just a single random byte the assertion will pass but still not correct.

I also don't know why I had to add 1 to `optsSubset.ToFileIdx` for the assertion to pass. Is this expected?

poszu: The comparison of the contents of the files, which overlap, is lower.

>I also don't know why I had to add 1 to optsSubset.ToFileIdx for the assertion to pass. Is this expected?

Because `ToFileIdx` is inclusive

cmd/postcli/main.go
poszu: Updated to be a method of `InitOpts`.

initialization/initialization_test.go
fasmat: Ah yes I see. Then we probably don't even need this check right? First reading the files and compare if one is contained in the other and then comparing the contents of the individual files again seems like overkill to me 🤷

fasmat (approved):

initialization/initialization_test.go
fasmat: I'm not blocking the merge because of this, just seems not necessary to me 🙂

initialization/initialization.go
```diff
@@ -325,22 +334,32 @@ func (init *Initializer) Initialize(ctx context.Context) error {
 	return fmt.Errorf("no nonce found")
 }
 
-func (init *Initializer) removeRedundantFiles(layout filesLayout) error {
-	numFiles, err := init.diskState.NumFilesWritten()
+func removeRedundantFiles(cfg config.Config, opts config.InitOpts, logger *zap.Logger) error {
```
fasmat, line 337: Why not leave it as method? you can still call it in tests if its private.

fasmat (approved):

initialization/initialization.go
poszu: To simplify testing - to avoid creating an instance of Initializer just to test this logic.

Issue #181, opened 2023-07-17
Verify POS initialization inflight


poszu:
We could catch if initialized labels are invalid very early on when they are created and before writing to a file. This would slow down initialization (how much depends on the % of labels checked), but give a higher confidence that the generated POS is valid and allow catching problems early.

It's essential that the problem of invalid labels is communicated to users in a way that is understandable so that they can take action by themself. We should not just `panic` with the message "labels are invalid" because users will be confused.

PR #183, opened 2023-07-17
Verify PoST labels against CPU provider while initialising

fasmat:
Part of #181 

During initialization for every batch that is calculated using the WorkOracle the first label will also be computed using the CPU provider. Should the labels not match return a typed `error` that can be used by `go-spacemesh`/`smapp` to display more detailed information to the user.


initialization/initialization.go
```diff
@@ -66,6 +60,7 @@ type option struct {
 
 	logger            *Logger
 	powDifficultyFunc func(uint64) []byte
+	referenceOracle   *oracle.WorkOracle
```
poszu, line 63: NIT, how about using this: https://github.com/spacemeshos/post/blob/3df31e0357396010460ef7f1ff0a0ccedbff27ff/internal/postrs/api.go#L34-L37
```suggestion
	referenceOracle   Scrypter
```

poszu (approved):

initialization/initialization.go
fasmat: `WorkOracle` doesn't implement `Scrypter`, instead it uses an `Scrypter` (implemented by `postrs.Scrypt`) and extends it with some additional error checking. Otherwise I would use it 🙂

PR #184, opened 2023-07-17
[postcli] verify initialized POS data

### LINKED ISSUE #180: Add POS verification to `postcli`
Add a new command to postcli to verify POS data at a given location. The idea is to cross-check a randomly picked subset of labels in the POS data against labels generated with the CPU implementation.

poszu:
Closes #180


cmd/postcli/main.go
```diff
@@ -198,3 +211,27 @@ func saveKey(key ed25519.PrivateKey) error {
...
+	params := postrs.TranslateScryptParams(opts.Scrypt.N, opts.Scrypt.R, opts.Scrypt.P)
+	verifyOpts := []postrs.VerifyPosOptionsFunc{
+		postrs.WithFraction(fraction),
+		postrs.FromFile(uint32(opts.FromFileIdx)),
+	}
+	if opts.ToFileIdx != nil {
+		verifyOpts = append(verifyOpts, postrs.ToFile(uint32(*opts.ToFileIdx)))
+	}
+
+	err := postrs.VerifyPos(opts.DataDir, params, verifyOpts...)
```
fasmat, line 288: Why not just return the error and if `err != nil` then `log.Fatal("failed to verify: %s", err)` ?

initialization/pos_verification_test.go
```diff
@@ -0,0 +1,60 @@
...
+	scryptParams := postrs.TranslateScryptParams(opts.Scrypt.N, opts.Scrypt.R, opts.Scrypt.P)
+
+	t.Run("valid", func(t *testing.T) {
+		err := postrs.VerifyPos(opts.DataDir, scryptParams, postrs.WithFraction(100.0))
+		require.NoError(t, err)
+	})
+	t.Run("invalid N", func(t *testing.T) {
+		wrongScrypt := postrs.TranslateScryptParams(4, opts.Scrypt.R, opts.Scrypt.P)
+		err := postrs.VerifyPos(opts.DataDir, wrongScrypt, postrs.WithFraction(100.0))
+		require.ErrorContains(t, err, "invalid POS")
```
fasmat:
```suggestion
		require.ErrorIs(t, err, postrs.ErrInvalidPos)
```

initialization/pos_verification_test.go
```diff
@@ -0,0 +1,60 @@
...
+	})
+	t.Run("corrupted data", func(t *testing.T) {
+		file, err := os.OpenFile(opts.DataDir+"/postdata_0.bin", os.O_WRONLY, 0)
+		require.NoError(t, err)
+		defer file.Close()
+		_, err = file.WriteAt([]byte("1234567890123456"), 0)
+		require.NoError(t, err)
+
+		err = postrs.VerifyPos(opts.DataDir, scryptParams, postrs.WithFraction(100.0))
+		require.ErrorContains(t, err, "invalid POS")
```
fasmat:
```suggestion
		require.ErrorIs(t, err, postrs.ErrInvalidPos)
```

internal/postrs/pos_verifier.go
```diff
@@ -0,0 +1,69 @@
...
+
+type VerifyPosOptions struct {
+	fromFile *uint32
+	toFile   *uint32
+	fraction float64
+}
+
+var ErrInvalidPos = errors.New("invalid POS")
+
+type VerifyPosOptionsFunc func(*VerifyPosOptions) error
```
fasmat, line 24: Could we also add `WithLogger` and make `setLogCallback` private again? Otherwise Pos Verification API would differ from all the others in this module in how to setup logging.

cmd/postcli/main.go
poszu: Didn't think of log.Fatal. Sure I can use it, but I still want to differentiate between "verification failure" and "verification successful but pos is invalid".

internal/postrs/pos_verifier.go
poszu: Hmm I'm not sure how to approach it best. `func WithLogger()` is already "taken" for opts in api.go. Should we start to divide into subpackages at this point?

fasmat: The quick way around it is to call it `WithVerifyPosLogger` or similar. I would like to avoid splitting the package just because of that.

Alternatively we could have a single `option` struct that is shared between this function and others in the package. Then we also only need one `WithLogger` functional option for all of them. Since this is an `internal` package I also think this would be OK, since it cannot be accidentally used incorrectly from another module.

pigmej:
@poszu we would need to also update readme to include that.

poszu:
@pigmej 
> @poszu we would need to also update readme to include that.

Yes, thanks for reminding me. Please use "request changes" in PR _Files changed_ tab 😉 . I added you to the reviewers to review the readme part when it's done. I'd really appreciate your input there.

Stizerg:
If it shows 'post is invalid' could it indicate in what file the error occurred. If it's just one file we could simply regenerate a faulty file.

pigmej (commented):
We really need to have *any* docs before we can consider deeper testing of it.
what does the `-fraction=1` mean? whole post? 1% of post?

Do I need to specify id and / or commitmentATX or it will read it from postdata_metadata?

poszu:
> If it shows 'post is invalid' could it indicate in what file the error occurred. If it's just one file we could simply regenerate a faulty file.

It already prints that:
```
2023-07-21T11:21:54.865+0200    INFO    POS data is invalid: invalid label in file 108 at offset 928
```

poszu:
> We really need to have _any_ docs before we can consider deeper testing of it. what does the `-fraction=1` mean? whole post? 1% of post?
> 
> Do I need to specify id and / or commitmentATX or it will read it from postdata_metadata?

@pigmej, updated the readme. Please check

cmd/postcli/README.md
```diff
@@ -158,3 +158,19 @@ in each `post_metadata.json` of every subset. Given two files:
 
 The nonce in the second file is the global minimum since its value is smaller than the first one. The operator is **required** to find the
 smallest VRF nonce by hand and ensure that its index and value are in the `postdata_metadata.json` of the merged directory on the target machine.
+
+
+## Verifying initialized POS data
+
+The `postcli` allows verifying an already initialized POS data. Verification randomly picks a small fraction of labels from every file and compares them to labels generated with the scrypt-jane algorithm executed on CPU. Please note that generating labels on CPU is slow, compared to GPU. Hence it is not possible to verify all the data (it would essentially mean re-initialization on CPU). The assumption is that if the GPU process breaks, it starts to output large chunks of invalid labels. This method will not detect a single invalid byte/label.
```
fasmat: Can you add a new line after 150 characters or so? Makes review suggestions easier 🙂

fasmat:
```suggestion
The `postcli` allows verifying an already initialized POS data. Verification samples a small fraction of labels from every file and compares them to labels generated with the same algorithm executed on CPU. Please note that generating labels on CPU is slow compared to GPU. Hence it is not possible to verify all the data (it would essentially mean re-initialization on CPU). If the GPU failed during initialization, the created PoST data will contain some or all invalid labels after that point. This method will only sample the PoST and might not detect a small amount of corrupted data.
```

cmd/postcli/README.md
```diff
@@ -158,3 +158,19 @@ in each `post_metadata.json` of every subset. Given two files:
...
+
+Depending on how much storage you have and your CPU speed you should pick the *fraction* (%) parameter that would give you enough confidence but still end in a reasonable time. Suggested values are <1%, closer to 0.1%.
```
fasmat:
```suggestion
Depending on PoST size and CPU speed a reasonable *fraction* (%) parameter needs to be picked that gives enough confidence but still completes verification in a reasonable time. Suggested values are <1%, closer to 0.1%.
```

cmd/postcli/README.md
```diff
@@ -158,3 +158,19 @@ in each `post_metadata.json` of every subset. Given two files:
...
+
+To verify POS data:
+
+1. locate the directory of the POS data. It should contain postdata_metadata.json and postdata_N.bin files.
+2. run `postcli -verify -datadir <path to POS directory> -fraction <% of data to verify>`.
+
+For example, `postcli -verify -datadir ~/post/data -fraction 0.1` will verify 0.1% of data. No additional arguments (i.e `-id`) are required. The postcli will read all required information from postdata_metadata.json
+
+If the POS data is found invalid, it will exit with 1 and print the index of invalid file and offset of the invalid label. Otherwise, it exits with 0.
```
fasmat:
```suggestion
If the POS data is found to be invalid, `postcli` will exit with status 1 and print the index of file and offset of the label found to be invalid. If verification completes successfully, `postcli` exits with 0.
```

pigmej (commented):
Looks good to me, @fasmat changes in grammar sounds good :)

Thx folks :)

pigmej (approved):

fasmat (approved):

haverlinkfor:
POS data is invalid: invalid label in file 108 at offset 928
after reinit of 108 file we again start from the first file. So we need to have an opportunity to start from 108

PR #186, opened 2023-07-17
Fix for release pipeline

### LINKED ISSUE #185: libpost is missing from postcli artifact
We must fix the pipeline so the needed libraries are included in the artifacts. Currently Linux build does not have `libpost.so` included

fasmat:
Closes https://github.com/spacemeshos/post/issues/185


poszu (commented):
The release artifacts are now somehow zipped twice. When I download an artifact it's "Windows.zip.zip".

poszu (approved):

PR #191, opened 2023-07-20
Update README.md and include in release

fasmat:
Made language in the README clearer as to what must be done by the operator.

README.md is now included in the release archive.

Additionally added a check that if a user initializes subsets of PoST they must provide an ID.


poszu (approved):
Let's wait for @pigmej approval before merging.

pigmej (approved):
Good additions! :)

PR #192, opened 2023-07-20
Improve logging in postcli

### LINKED ISSUE #188: postcli limits log messages to 100/s
The postcli is configured with very opinionated zap's _Production_ preset. It is configured with a `Sampler` limiting the rate of logging up to 100 logs/s. It causes big confusion among users - postcli output looks like it was broken.

poszu:
Closes #188

### DIFF
--- a/cmd/postcli/main.go
+++ b/cmd/postcli/main.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 
 	"github.com/spacemeshos/post/config"
 	"github.com/spacemeshos/post/initialization"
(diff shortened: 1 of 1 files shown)


cmd/postcli/main.go
```diff
@@ -121,15 +126,32 @@ func main() {
...
+			LineEnding:     zapcore.DefaultLineEnding,
+			EncodeLevel:    zapcore.CapitalLevelEncoder,
+			EncodeTime:     zapcore.ISO8601TimeEncoder,
+			EncodeDuration: zapcore.StringDurationEncoder,
+		},
+		OutputPaths:      []string{"stdout"},
+		ErrorOutputPaths: []string{"stderr"},
+	}
+
+	zapLog, err := zapCfg.Build()
```
fasmat, line 146: I assume the goal is to have human readable logs with the specified level? Would this work?

```go
	zapCfg := zap.NewDevelopmentConfig()
	zapCfg.Level = zap.NewAtomicLevelAt(logLevel)
	zapLog, err := zapCfg.Build()
```

fasmat (approved):

cmd/postcli/main.go
poszu: It's actually a slightly tweaked Development preset with a few changes:
- don't print the file and line where a log was emitted,
- don't print stack on ERROR level log,
- print to stdout instead of stderr

PR #193, opened 2023-07-20
[postcli] option to look for lost VRF nonce in initialized data

### LINKED ISSUE #190: Add functionality to look for a VRF nonce in already initialized data

poszu:
Closes #190 

- [ ] need to see if doing it single-threaded is sufficiently fast.

### DIFF
--- a/initialization/vrf_search.go
+++ b/initialization/vrf_search.go
@@ -95,6 +95,12 @@ func SearchForNonce(ctx context.Context, cfg Config, initOpts InitOpts, opts ...
 		if err != nil {
 			logger.Panic("failed to parse file index", zap.Error(err))
 		}
+
+		if fileIndex < initOpts.FromFileIdx || (initOpts.ToFileIdx != nil && fileIndex > *initOpts.ToFileIdx) {
+			logger.Debug("skipping file", zap.String("file", file.Name()))
+			continue
+		}
+
 		firstLabelIndex := uint64(fileIndex) * metadata.MaxFileSize / postrs.LabelLength
 		logger.Info("looking for VRF nonce in file", zap.String("file", file.Name()))
 
--- a/initialization/vrf_search.go
+++ b/initialization/vrf_search.go
@@ -111,36 +111,48 @@ func SearchForNonce(ctx context.Context, cfg Config, initOpts InitOpts, opts ...
 		}
 		defer file.Close()
 
-		idx, l, err := searchForNonce(ctx, bufio.NewReader(file), difficulty, woReference)
-		if err != nil {
-			return 0, nil, fmt.Errorf("failed to search for nonce: %w", err)
-		}
-		if l != nil {
-			nonceValue = l
+		idx, label, err := searchForNonce(ctx, bufio.NewReader(file), difficulty, woReference)
+		if label != nil {
+			nonceValue = label
 			nonce = firstLabelIndex + idx
-			logger.Info("found nonce", zap.Uint64("index", nonce), zap.String("label", hex.EncodeToString(nonceValue)))
+			if err := persistNonce(nonce, nonceValue, metadata, initOpts.DataDir, logger); err != nil {
+				return nonce, nonceValue, err
+			}
 			difficulty = nonceValue // override difficulty to the new lowest label
+		}
 
+		switch {
+		case errors.Is(err, context.Canceled):
+			logger.Info("search for nonce interrupted", zap.Uint64("nonce", nonce), zap.String("nonceValue", hex.EncodeToString(nonceValue)))
+			return nonce, nonceValue, err
+		case err != nil:
+			return 0, nil, fmt.Errorf("failed to search for nonce: %w", err)
 		}
 	}
 	if nonceValue != nil {
-		logger.Info("updating postdata_metadata.json with found nonce", zap.Uint64("nonce", nonce), zap.String("NonceValue", hex.EncodeToString(nonceValue)))
-		metadata.Nonce = &nonce
-		metadata.NonceValue = shared.NonceValue(nonceValue)
-		SaveMetadata(initOpts.DataDir, metadata)
 		return nonce, nonceValue, nil
 	}
 	return 0, nil, ErrNonceNotFound
 }
 
+func persistNonce(nonce uint64, label []byte, metadata *shared.PostMetadata, datadir string, logger *zap.Logger) error {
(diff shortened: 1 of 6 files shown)


initialization/vrf_search.go
```diff
@@ -0,0 +1,183 @@
...
+		default:
+			// continue looking for a nonce
+		}
+
+		_, err := io.ReadFull(r, labelBuf)
+		switch {
+		case err == io.EOF:
+			return nonce, nonceValue, nil
+		case err == io.ErrUnexpectedEOF:
+			return 0, nil, errors.New("unexpected end of file - file is truncated. please reinit it")
```
fasmat:
```suggestion
			return 0, nil, fmt.Errorf("file appears truncated. please re-init it: %w", err)
```

initialization/vrf_search.go
```diff
@@ -0,0 +1,183 @@
...
+		return true, nil
+	} else if comp == 0 {
+		// need to regenerate label to verify lower bytes
+		res, err := wo.Position(index)
+		if err != nil {
+			return false, fmt.Errorf("failed to regenerate label: %w", err)
+		}
+		return bytes.Compare(res.Output, difficulty) < 0, nil
+	}
+	return false, nil
```
fasmat: Would it make sense to use switch here?

```suggestion
	switch {
	case comp < 0:
		return true, nil
	case comp == 0:
		// need to regenerate label to verify lower bytes
		res, err := wo.Position(index)
		if err != nil {
			return false, fmt.Errorf("failed to regenerate label: %w", err)
		}
		return bytes.Compare(res.Output, difficulty) < 0, nil
	default: // comp > 0
		return false, nil
	}
```

initialization/vrf_search.go
```diff
@@ -0,0 +1,183 @@
...
+		return nonce, nonceValue, nil
+	}
+	return 0, nil, ErrNonceNotFound
+}
+
+// searchForNonce searches for a nonce in the given reader.
+func searchForNonce(ctx context.Context, r io.Reader, difficulty []byte, oracle *oracle.WorkOracle) (nonce uint64, nonceValue []byte, err error) {
+	labelBuf := make([]byte, postrs.LabelLength)
+	labelIndex := uint64(0)
+	for {
```
fasmat:
```suggestion
	labelIndex := uint64(0)
	for labelIndex := uint64(0); ; labelIndex++ {
```

initialization/vrf_search.go
```diff
@@ -0,0 +1,183 @@
...
+			return 0, nil, errors.New("unexpected end of file - file is truncated. please reinit it")
+		}
+
+		ok, err := checkLabel(labelIndex, labelBuf, difficulty, oracle)
+		if err != nil {
+			return 0, nil, fmt.Errorf("checking label: %w", err)
+		}
+		if ok {
+			nonce = labelIndex
+			nonceValue = append(nonceValue[:0], labelBuf...)
```
fasmat, line 180:
```suggestion
			copy(nonceValue, labelBuf)
```

initialization/vrf_search.go
```diff
@@ -0,0 +1,183 @@
...
+			difficulty = nonceValue // override difficulty to the new lowest label
```
fasmat, line 181: Maybe also `copy` just to be safe?

initialization/vrf_search.go
```diff
@@ -0,0 +1,183 @@
...
+
+		idx, l, err := searchForNonce(ctx, bufio.NewReader(file), difficulty, woReference)
+		if err != nil {
+			return 0, nil, fmt.Errorf("failed to search for nonce: %w", err)
+		}
+		if l != nil {
+			nonceValue = l
+			nonce = firstLabelIndex + idx
+			logger.Info("found nonce", zap.Uint64("index", nonce), zap.String("label", hex.EncodeToString(nonceValue)))
+			difficulty = nonceValue // override difficulty to the new lowest label
```
fasmat, line 127: maybe `copy` just to be safe?

initialization/vrf_search.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+}
+
+// checkLabels checks if label is lower than difficulty.
+// It will regenerate the whole 32B of the label if its most significant 16B are equal to difficulty
+// in order to check the lower bytes.
+// * index:      the index of the label (used to regenerate it)
+// * label:		 16B most significant bytes of the label
+// * difficulty: all 32b of the difficulty
+// * wo:         the oracle used to regenerate the label.
```
fasmat: Formatting appears to be off?

initialization/vrf_search.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// searchForNonce searches for a nonce in the given reader.
+func searchForNonce(ctx context.Context, r io.Reader, difficulty []byte, oracle *oracle.WorkOracle) (nonce uint64, nonceValue []byte, err error) {
+	labelBuf := make([]byte, postrs.LabelLength)
+	labelIndex := uint64(0)
+	for {
+		select {
+		case <-ctx.Done():
+			return nonce, labelBuf, ctx.Err()
```
fasmat: The way it is implemented at the moment if once ctrl+C the search after a nonce was found it won't be saved. Only after the search is completed.

Intended?

fasmat:
> * [ ]  need to see if doing it single-threaded is sufficiently fast.

I'm 99% sure the bottleneck is the read speed from disk in most cases at least that's how we designed PoST 🙂

poszu:
> > * [ ]   need to see if doing it single-threaded is sufficiently fast.
> 
> I'm 99% sure the bottleneck is the read speed from disk in most cases at least that's how we designed PoST 🙂

Yes, I also think so.

initialization/vrf_search.go
poszu: I use append to copy because initially `nonceValue` is `nil`.

initialization/vrf_search.go
poszu: Do you mean to not modify `difficulty` by mistake by modifying nonceValue?

initialization/vrf_search.go
poszu: Thanks, good catch :+1:

poszu:
@pigmej would you like to review the documentation?

pigmej (approved):
Good docs addition :)

initialization/vrf_search.go
fasmat: yes. Just as a safety measure, in terms of performance it should make little to no difference.

fasmat (approved):

PR #195, opened 2023-07-20
Improve VRF lookup past init data

### LINKED ISSUE #187: Disable looking for VRF nonce past initialized data when using `-from/toFile` flag
The postcli can be used to initialize only a subset of data using `-fromFile` and `-toFile` flags (for example when re-initializing corrupted/missing data **on a different machine** or splitting initialization between machines). 

The current behavior is to continue looking for VRF nonce (which is very likely to happen if initializing only a subset). It should be possible to disable looking for VRF nonce past the requested size (either automatically when `-fromFile` or `-toFile`flag is used or via a flag like `-disableEagerVrfSearch` or something like that.

fasmat:
Closes #187 

No additional command line flags are required. The initializer will never look outside the range it initializes if it was instructed to initialize less than the whole range.

### DIFF
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -309,6 +309,11 @@ func (init *Initializer) Initialize(ctx context.Context) error {
 	}
 
 	if init.nonce.Load() != nil {
+		init.logger.Info("initialization: completed, found nonce", zap.Uint64("nonce", *init.nonce.Load()))
+		return nil
+	}
+	if layout.FirstFileIdx > 0 || layout.LastFileIdx < init.opts.TotalFiles(init.cfg.LabelsPerUnit)-1 {
+		init.logger.Info("initialization: no nonce found while computing labels")
 		return nil
 	}
 
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -255,14 +255,11 @@ func (init *Initializer) Initialize(ctx context.Context) error {
 		zap.Uint64("labelsPerUnit", init.cfg.LabelsPerUnit),
 	)
 
-	lastFileIndex := layout.FirstFileIdx + int(layout.NumFiles) - 1
-
 	init.logger.Info("initialization file layout",
-		zap.Uint("numFiles", layout.NumFiles),
 		zap.Uint64("labelsPerFile", layout.FileNumLabels),
 		zap.Uint64("labelsLastFile", layout.LastFileNumLabels),
 		zap.Int("firstFileIndex", layout.FirstFileIdx),
-		zap.Int("lastFileIndex", lastFileIndex),
+		zap.Int("lastFileIndex", layout.LastFileIdx),
 	)
 	if err := removeRedundantFiles(init.cfg, init.opts, init.logger); err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


initialization/initialization_test.go
```diff
@@ -934,46 +939,84 @@ func TestInitializeSubset(t *testing.T) {
...
 
 	// Verify that postdata_3.bin from both initializations contain the same data
 	fullPostdata3, err := os.ReadFile(filepath.Join(opts.DataDir, "postdata_3.bin"))
-	r.NoError(err)
+	require.NoError(t, err)
 	subsetPostdata3, err := os.ReadFile(filepath.Join(optsSubset.DataDir, "postdata_3.bin"))
-	r.NoError(err)
-	r.Equal(fullPostdata3, subsetPostdata3)
+	require.NoError(t, err)
+	require.Equal(t, fullPostdata3, subsetPostdata3)
```
poszu, line 971: Why these changes? IMO it doesn't make sense to change back and forth between the two. Let's establish one guideline and keep it everywhere.

initialization/initialization_test.go
```diff
@@ -934,46 +939,84 @@ func TestInitializeSubset(t *testing.T) {
...
+	cfg := config.DefaultConfig()
+	cfg.LabelsPerUnit = 128
+
+	opts := config.DefaultInitOpts()
+	opts.DataDir = t.TempDir()
+	opts.FromFileIdx = 3
+	opts.ToFileIdx = new(int)
+	*opts.ToFileIdx = 4
+	opts.NumUnits = 20
+	opts.MaxFileSize = cfg.LabelsPerUnit * 2 * uint64(config.BytesPerLabel()) // 2 units per file
```
poszu:
```suggestion
	opts.MaxFileSize = cfg.UnitSize()* 2
```

initialization/initialization_test.go
```diff
@@ -934,46 +939,84 @@ func TestInitializeSubset(t *testing.T) {
...
+		WithInitOpts(opts),
+		WithLogger(zaptest.NewLogger(t, zaptest.Level(zap.DebugLevel))),
+		// use a higher difficulty to make sure no Pow is found in the first `numLabels` labels.
+		withDifficultyFunc(func(numLabels uint64) []byte {
+			x := new(big.Int).Lsh(big.NewInt(1), 256)
+			x.Div(x, big.NewInt(int64(numLabels)))
+
+			difficulty := make([]byte, 32)
+			return x.FillBytes(difficulty)
+		}),
```
poszu, line 1008: This should work as well
```suggestion
		withDifficultyFunc(func(_ uint64) []byte {
			return make([]byte, 32)
		}),
```

initialization/initialization.go
```diff
@@ -312,6 +309,11 @@ func (init *Initializer) Initialize(ctx context.Context) error {
 	}
 
 	if init.nonce.Load() != nil {
+		init.logger.Info("initialization: completed, found nonce", zap.Uint64("nonce", *init.nonce.Load()))
+		return nil
+	}
+	if layout.FirstFileIdx > 0 || layout.LastFileIdx < init.opts.TotalFiles(init.cfg.LabelsPerUnit)-1 {
+		init.logger.Info("initialization: no nonce found while computing labels")
 		return nil
 	}
```
poszu: How about adding a method to layout struct:
```go
func (l filesLayout) NumFiles() int {
    return l.LasFileIdx - l.FirstFileIdx + 1
}
```

the code would become more readable I think:
```suggestion
	if layout.NumFiles() < init.opts.TotalFiles(init.cfg.LabelsPerUnit) {
		init.logger.Info("initialization: no nonce found while computing labels")
		return nil
	}
```

poszu (approved):
LGTM in general, left few nits

initialization/initialization_test.go
fasmat: Yeah, I only changed it in this test, because this test was using both options and I used it as a template for the new test.

initialization/initialization_test.go
fasmat: I actually had a reason to use just a more difficult function than an impossible one. I extended the test to see if after we initialized a subset without nonce, initializing the full PoST finds the nonce (also outside the initialized range).

PR #196, opened 2023-07-20
Prevent possibly accidental overwrite of key.bin

fasmat:
Quick fix to prevent a possibly major mistake


pigmej (commented):
Maybe just better error msg? Something more obvious that would indicate "if you really want to continue delete file by hand" or something like that.

pigmej (approved):

PR #205, opened 2023-07-31
Verify key.bin against NodeID in metadata

### LINKED ISSUE #198: Postcli could verify key.bin vs NodeID in postdata_metatada.json
The `-verify` option in postcli currently only verifies POS data against information read from postdata_metadata.json. It could also verify if the public key in key.bin matches nodeID in postdata_metadata.json

fasmat:
Closes #198.

Simple verification of the private key in `key.bin` against the NodeID in `postdata_metadata.json`. Will be done when `-verify` is set.

### DIFF
--- a/cmd/postcli/main.go
+++ b/cmd/postcli/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"crypto/ed25519"
 	"encoding/hex"
@@ -275,6 +276,36 @@ func saveKey(key ed25519.PrivateKey) error {
 }
 
 func cmdVerifyPos(opts config.InitOpts, fraction float64, logger *zap.Logger) error {
+	log.Println("cli: verifying key.bin")
+
+	filename := filepath.Join(opts.DataDir, edKeyFileName)
+	data, err := os.ReadFile(filename)
+	if err != nil {
+		return fmt.Errorf("key read from disk error: %w", err)
+	}
+
+	dst := make([]byte, ed25519.PrivateKeySize)
+	n, err := hex.Decode(dst, data)
+	if err != nil {
+		return fmt.Errorf("decoding private key: %w", err)
+	}
+	if n != ed25519.PrivateKeySize {
+		return fmt.Errorf("invalid key size %d, expected %d", n, ed25519.PrivateKeySize)
+	}
+	pub := ed25519.NewKeyFromSeed(dst).Public().(ed25519.PublicKey)
+
+	meta, err := initialization.LoadMetadata(opts.DataDir)
+	if err != nil {
+		return fmt.Errorf("failed to load metadata: %w", err)
+	}
+
+	if !bytes.Equal(meta.NodeId, pub) {
+		return fmt.Errorf("failed to verify node id (%x) against key.bin", meta.NodeId)
+	}
+
+	log.Println("cli: key.bin is valid")
+	log.Println("cli: verifying POS data")
+
 	params := postrs.TranslateScryptParams(opts.Scrypt.N, opts.Scrypt.R, opts.Scrypt.P)
 	verifyOpts := []postrs.VerifyPosOptionsFunc{
 		postrs.WithFraction(fraction),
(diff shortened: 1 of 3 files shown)


cmd/postcli/main.go
```diff
@@ -275,6 +276,36 @@ func saveKey(key ed25519.PrivateKey) error {
...
+	}
+	pub := ed25519.NewKeyFromSeed(dst).Public().(ed25519.PublicKey)
+
+	meta, err := initialization.LoadMetadata(opts.DataDir)
+	if err != nil {
+		return fmt.Errorf("failed to load metadata: %w", err)
+	}
+
+	if !bytes.Equal(meta.NodeId, pub) {
+		return fmt.Errorf("failed to verify node id (%x) against key.bin", meta.NodeId)
```
poszu: This error message could be confusing. It didn't _fail to verify_ - it verified successfully but detected a problem. How about wording it differently so that it's obvious to users **and** helping them troubleshoot by themselves? Something like:
```suggestion
		return fmt.Errorf("NodeID in %s doesn't match public key in %s (%x != %x)", filepath.Join(opts.DataDir, initialization.MetadataFileName), filename, meta.NodeId, pub)
```

(export `const MetadataFileName` in metadata.go)

cmd/postcli/main.go
```diff
@@ -275,6 +276,36 @@ func saveKey(key ed25519.PrivateKey) error {
 }
 
 func cmdVerifyPos(opts config.InitOpts, fraction float64, logger *zap.Logger) error {
+	log.Println("cli: verifying key.bin")
+
+	filename := filepath.Join(opts.DataDir, edKeyFileName)
+	data, err := os.ReadFile(filename)
+	if err != nil {
+		return fmt.Errorf("key read from disk error: %w", err)
```
poszu: ditto:
```suggestion
		return fmt.Errorf("could not read keys from %s: %w", filename, err)
```

... unless `err` already contains the path.

cmd/postcli/main.go
poszu: Note that the returned error is not logged. The main function simply `os.Exit(1)`s on error.

fasmat: Errors are not print statements though. I updated to directly log the issue.

fasmat: Changed code to log instead of returning an error

cmd/postcli/main.go
```diff
@@ -167,10 +167,7 @@ func main() {
 	}
 
 	if verifyPos {
-		if cmdVerifyPos(opts, fraction, logger) != nil {
-			os.Exit(1)
-		}
-		os.Exit(0)
+		cmdVerifyPos(opts, fraction, logger)
```
poszu: Previously it was clear that the program terminates after calling `cmdVerifyPos()`. Now it's hidden and less obvious. How about returning an `error` and adding a `log.Fatal()` here?

fasmat: I don't see the benefit? Errors are not logging messages. I changed the code so it is clear it exits after `cmdVerifyPos` is called.

poszu: > Errors are not logging messages.

? The point of CLI is to communicate to the user the outcome. What's wrong with communicating by showing them a nice readable error message?

>I don't see the benefit?

The benefit is that the next person reading this code immediately understands that the program terminates here.

cmd/postcli/main.go
poszu: I don't know what you mean. Good errors contain enough information to troubleshoot the problem, not just "oops failed".

poszu (approved):

cmd/postcli/main.go
fasmat: > What's wrong with communicating by showing them a nice readable error message?

I did this by updating the errors with your suggestions and printing them directly? Why first wrap them in an error to return them just to print them?

> The benefit is that the next person reading this code immediately understands that the program terminates here.

That's why I added the `os.Exit(0)` as you suggested?

xx-134:
Ashampoo_Snap_31 июля 2023 г _15h20m22s_156_
What am I doing wrong?

fasmat:
@xx-134 thanks for reporting. This was an oversight that will be fixed in this PR: https://github.com/spacemeshos/post/pull/207

Issue #199, opened 2023-07-21
Introduce cobra for postcli


fasmat:
The code in `postcli` is now complex enough that it becomes hard to follow which flags are intended for which commands and how everything fits together.

- This needs to be cleaned up preferably with a library like [cobra](https://github.com/spf13/cobra).
- Existing code should be split into sub commands (`init`, `verify`, `resize`, `bench`, etc.)
- The sub commands should have some rudimentary tests
- A new sub command `version` should be added that prints the version of `postcli`/`post`.

PR #207, opened 2023-07-31
Fix incorrect seed length during verify

fasmat:
In #205 the whole private key is passed to `ed25519.NewKeyFromSeed` while it should only be the first 32 bytes (the seed - the rest is the public key).

This PR fixes this oversight.


poszu (approved):
👍 Perhaps postcli deserves some unit tests.

fasmat:
> 👍 Perhaps postcli deserves some unit tests.

Will come with https://github.com/spacemeshos/post/issues/199

PR #208, opened 2023-07-31
Avoid initializing work oracle if not needed

poszu:
Currently, the work oracle (de facto the OpenCL provider) is created always, regardless of whether labels need to be generated. This is suboptimal and causes issues for users who moved their data to a setup w/o a GPU OpenCL provider (but forgot to remove provider-id configuration).

The proposed solution is to initialize the work oracle lazily, only when it is needed.

### DIFF
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -159,6 +159,21 @@ func withReferenceOracle(referenceOracle *oracle.WorkOracle) OptionFunc {
 	}
 }
 
+// Lazy initialized value of type T.
+type LazyInit[T any] struct {
+	init     func() (T, error)
+	initOnce sync.Once
+	value    T
+	err      error
+}
+
+func (l *LazyInit[T]) Get() (T, error) {
+	l.initOnce.Do(func() {
+		l.value, l.err = l.init()
+	})
+	return l.value, l.err
+}
+
 // Initializer is responsible for initializing a new PoST commitment.
 type Initializer struct {
 	nodeId          []byte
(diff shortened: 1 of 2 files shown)


initialization/initialization.go
```diff
@@ -302,31 +317,42 @@ func (init *Initializer) Initialize(ctx context.Context) error {
...
+				oracle.WithScryptParams(init.opts.Scrypt),
+				oracle.WithLogger(init.logger),
+			)
+		}}
+
+		defer func() {
+			if woReference.value != nil {
+				woReference.value.Close()
+			}
+		}()
```
fasmat: I see this as problematic, if `woReference` changes between deferring the function and it being invoked, it won't call `Close` on the right object.

initialization/initialization.go
```diff
@@ -159,6 +159,21 @@ func withReferenceOracle(referenceOracle *oracle.WorkOracle) OptionFunc {
...
+	err      error
+}
+
+func (l *LazyInit[T]) Get() (T, error) {
+	l.initOnce.Do(func() {
+		l.value, l.err = l.init()
+	})
+	return l.value, l.err
+}
+
```
fasmat: I don't see a need to do this in a generic way. I doubt this code will be reused and if it will its small and simple enough that copying it to where its needed is easier than having it in some new package to be re-used 🤷 

Making it implement `io.Closer` also becomes easier if its specific rather than generic:

```go
type LazyOracle struct {
	init  func() (*oracle.WorkOracle, error)
	once  sync.Once
	value *oracle.WorkOracle
	err   error
}

func (l *LazyOracle) Get() (T, error) {
	l.once.Do(func() {
		l.value, l.err = l.init()
	})
	return l.value, l.err
}

func (l *lazyOracle) Close() (error) {
	if l.value != nil {
		return l.value.Close()
	}
}
```

fasmat (commented):
I personally think it would be better to make the `WorkOracle` lazy and instantiate the underlying `postrs.Scrypt` on the first call to `oracle.Positions` instead of doing this in the initializer.

fasmat (approved):

Issue #212, opened 2023-08-04
Remove automatic selection for best provider


fasmat:
### Description

At the moment the node defaults to provider -1 in their config if not set to a different provider by the user. This value is problematic for multiple reasons:

1. The provider ID is an `uint` which means `-1` will be cast to either `2^32 - 1` or `2^64 - 1` depending on the system.
2. The CPU provider of post-rs has the hard coded ID `2^32 -1` which can cause the node to do a benchmarks when not intended.

### Acceptance criteria

- Remove `-1` as identifier for "select best provider". The user has to select a provider explicitly via Smapp or the config file.
- Remove automatic benchmarking; but not manually triggered (used by Smapp to display speeds of providers to user)
- Default to CPU provider in config, so that unit and system tests still behave as expected

PR #213, opened 2023-08-04
Remove best provider and make ProviderID mandatory in InitOpts

fasmat:
Part of https://github.com/spacemeshos/post/issues/212

- Removes `BestProviderID`
- Change ProviderID from `uint` to `*uint32`
- Removes default ProviderID in `DefaultInitOpts`
- Makes selecting a provider ID in `postcli` mandatory (before defaulted to CPU)

### DIFF
--- a/initialization/initialization_test.go
+++ b/initialization/initialization_test.go
@@ -811,7 +811,7 @@ func TestValidateComputeBatchSize(t *testing.T) {
 func TestWrongLabelsDetected(t *testing.T) {
 	cfg, opts := getTestConfig(t)
 
-	logger := zaptest.NewLogger(t, zaptest.Level(zap.DebugLevel))
+	logger := zaptest.NewLogger(t)
 
 	woReference, err := oracle.New(
 		oracle.WithCommitment(make([]byte, 32)), // different commitment to trigger error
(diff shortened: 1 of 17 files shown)


oracle/oracle.go
```diff
@@ -52,11 +52,10 @@ func (o *option) validate() error {
...
-// WithProviderID sets the ID of the openCL provider to use.
-func WithProviderID(id uint) OptionFunc {
+// WithProviderID sets the ID of the OpenCL provider to use.
+func WithProviderID(id *uint32) OptionFunc {
 	return func(opts *option) error {
-		opts.providerID = new(uint)
-		*opts.providerID = id
+		opts.providerID = id
 		return nil
 	}
```
poszu, line 57: IMO, accepting just `uint32` would make a nicer API for two reasons:
- can use the output of a function directly (i.e. `WithProviderID(CpuProviderId())`)
- cannot pass an invalid `nil` value at all

```go
func WithProviderID(id uint32) OptionFunc {
	return func(opts *option) error {
		opts.providerID = &id
		return nil
	}
```

initialization/initialization_test.go
```diff
@@ -30,15 +29,21 @@ var (
...
 )
 
-func TestInitialize(t *testing.T) {
+func getTestConfig(tb testing.TB) (config.Config, config.InitOpts) {
 	cfg := config.DefaultConfig()
-	cfg.LabelsPerUnit = 1 << 12
 
 	opts := config.DefaultInitOpts()
-	opts.DataDir = t.TempDir()
+	opts.Scrypt.N = 16 // speed up initialization
```
poszu, line 36: This could actually be the smallest possible value of 2 for best test speed
```suggestion
	opts.Scrypt.N = 2 // speed up initialization
```

💡 IIRC, there was a test that didn't want to work with N=2 because... init was too fast (a test checking initializer state transitions or something like that).

poszu (commented):
It would be good to verify if go-spacemesh works fine with this change.

initialization/initialization_test.go
fasmat: With `N=2` at least one test breaks and with `N=16` all tests in the `initialization` package combined take ~ 6 seconds to complete on `ubuntu-latest`.

oracle/oracle.go
fasmat: Using an option for ID here, kinda goes against the goal of the implementation. That means that `nil` cannot be passed any more to the WorkOracle inside the `Initializer` service and the node has to provide some (even if invalid) provider ID even if not needed.

fasmat: Specifically this section here: https://github.com/spacemeshos/post/blob/7e32021db1c1a842f9dafec159e78c1a9d8a4dd7/initialization/initialization.go#L306-L312

fasmat: I went with the following: `oracle.WithProviderID`takes a pointer to the provider ID that can be `nil` and will be checked for as soon as the underlying `postrs.Scrypter` is lazily initialized. Which will return an error that can be propagated upwards and logged.

This way if no labels need to be generated the provider can be `nil` without triggering an error.

poszu (approved):

PR #215, opened 2023-08-09
Change logging level of message to avoid confusion.

fasmat:
A simple temporary workaround for the issue https://github.com/spacemeshos/go-spacemesh/issues/4789

This changes the level of logging messages that report unrecognised files from WARN to DEBUG.

PoST ignores files it doesn't know. At the moment the node creates additional files in the data directory. PoST reporting them as warning to users might mislead them into believing these files don't belong there or shall be deleted. Only showing these messages when debug logging is active should prevent this.

### DIFF
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -410,6 +410,7 @@ func removeRedundantFiles(cfg config.Config, opts config.InitOpts, logger *zap.L
 		name := file.Name()
 		fileIndex, err := shared.ParseFileIndex(name)
 		if err != nil && name != MetadataFileName {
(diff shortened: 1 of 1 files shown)


poszu:
I'd not call it a fix but a workaround for a problem that has its source in go-spacemesh. These files should not be there, hence a warning. Could you add a _TODO_ to bring back the warning level after go-sm is refactored to not store files in the same dir as POS?

fasmat:
I added a TODO and updated the PR description to workaround instead of fix 🙂

poszu (approved):

PR #222, opened 2023-08-17
Add -threads and -nonces arguments to be used with -genproof

### LINKED ISSUE #221: add -theads and -nonces into postcli -genproof
To have a more realistic time and `-genproof` we should add threads and nonces to the CLI.
Also that will mean that the `-genproof` will be able to do exactly the same as node will do in cycle_gap.

chrisxch:
Closes https://github.com/spacemeshos/post/issues/221

### DIFF
--- a/cmd/postcli/main.go
+++ b/cmd/postcli/main.go
@@ -47,6 +47,9 @@ var (
 	commitmentAtxId    []byte
 	reset              bool
 
+	nonces	uint
+	threads	uint
+
 	logLevel zapcore.Level
 
 	ErrKeyFileExists = errors.New("key file already exists")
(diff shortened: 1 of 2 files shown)


cmd/postcli/main.go
```diff
@@ -74,6 +77,9 @@ func parseFlags() {
 	flag.StringVar(&commitmentAtxIdHex, "commitmentAtxId", "", "commitment atx id, in hex (required)")
 	numUnits := flag.Uint64("numUnits", uint64(opts.NumUnits), "number of units")
 
+	flag.UintVar(&nonces, "nonces", 16, "number of nonces for -genproof")
```
poszu, line 80: The number of nonces must be a multiple of 16 (a nonce group). How about changing this argument into "number of nonce groups" to avoid setting an invalid value? Also, 16 is unrealistic, perhaps a more typical value like `288` would be better?
```suggestion
	flag.UintVar(&noncesGroups, "nonceGroups", 18, "number of nonce groups for -genproof. Each group consists of 16 nonces.")
```
And multiply it by 16 later in the code.

OR at least inform the user that it must be a multiple of 16:
```suggestion
	flag.UintVar(&nonces, "nonces", 16, "number of nonces for -genproof. Must be a multiple of 16.")
```

proving/proving.go
```diff
@@ -34,6 +35,8 @@ func Generate(ctx context.Context, ch shared.Challenge, cfg config.Config, logge
 		provingOpts = append(provingOpts, postrs.WithPowCreator(options.powCreatorId))
 	}
 
+	log.Printf("Generating proof with: %d nonces and %d threads.\n", options.nonces, options.threads)
+	logger.Info("proving: start generating proof")
```
poszu, line 39: Nonces and threads can be added directly into the second log.
```suggestion
	logger.Info("proving: start generating proof", zap.Uint("nonces", options.nonces), zap.Uint("threads", options.threads))
```

fasmat:
Closing due to inactivity, feel free to re-open if you think this is still useful 🙂

PR #224, opened 2023-08-18
feat: support get metadata dir from env

maxvint:
If you use some distributed storage systems, may encounter situations where you cannot read and save `postdata_metadata.json`, and an option to support get the metadata directory from environment variable.

### DIFF
--- a/initialization/metadata.go
+++ b/initialization/metadata.go
@@ -11,7 +11,17 @@ import (
 
 const MetadataFileName = "postdata_metadata.json"
 
+func getEffectiveDir(dir string) string {
+	envDir := os.Getenv("POST_METADATA_DIR")
+	if envDir != "" {
+		return envDir
+	}
+	return dir
+}
+
 func SaveMetadata(dir string, v *shared.PostMetadata) error {
+	dir = getEffectiveDir(dir)
+
 	err := os.MkdirAll(dir, shared.OwnerReadWriteExec)
 	if err != nil && !os.IsExist(err) {
 		return fmt.Errorf("dir creation failure: %w", err)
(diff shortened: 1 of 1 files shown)


poszu:
@maxvint, thanks for your contribution. Could you explain what is the potential problem with distributed storage systems? Is it a problem with the directory being read-only?

initialization/metadata.go
```diff
@@ -22,7 +33,11 @@ func SaveMetadata(dir string, v *shared.PostMetadata) error {
 		return fmt.Errorf("serialization failure: %w", err)
 	}
 
+	filePath := filepath.Join(dir, MetadataFileName)
+	fmt.Println("Writing metadata to:", filePath)
```
poszu, line 37: Please, avoid using prints. You can log with loggers.

initialization/metadata.go
```diff
@@ -11,7 +11,18 @@ import (
 
 const MetadataFileName = "postdata_metadata.json"
 
+func getEffectiveDir(dir string) string {
+	envDir := os.Getenv("POST_METADATA_DIR")
+	fmt.Println("Got postmeta dir of env:", envDir)
+	if envDir != "" {
+		return envDir
+	}
+	return dir
```
poszu, line 20: How about using `os.LookupEnv` instead:
```suggestion
	if value, ok := os.LookupEnv("POST_METADATA_DIR"); ok {
        	return value
	}
	return dir
```

initialization/metadata.go
```diff
@@ -11,7 +11,18 @@ import (
 
 const MetadataFileName = "postdata_metadata.json"
 
+func getEffectiveDir(dir string) string {
```
poszu, line 14: It would be nice to cover it with a unit test.

poszu (changes requested):

poszu:
Closing as stale. @maxvint please reopen if you still wish to work on it and merge.

PR #225, opened 2023-08-18
Check for curl version when downloading libs

fasmat:


Makefile.Inc
```diff
@@ -29,6 +29,7 @@ ifeq ($(GOOS),windows)
 	platform := windows
 	export PATH := $(PATH):$(PROJ_DIR)build
 	EXE := .exe
+	CGO_LDFLAGS := $(CGO_LDFLAGS) -Wl,-Bstatic -lpthread -Wl,-Bdynamic
```
poszu, line 32: Why this change?

poszu (approved):

Makefile.Inc
fasmat: To align with the Makefile in `go-spacemesh`, otherwise we are building `postcli` with different `LDFLAGS` than `go-spacemesh`.

The change that introduced these flags for windows was: https://github.com/spacemeshos/go-spacemesh/pull/4581

Seems like no one has been using `postcli` on windows yet, or we would have already seen complaints about missing dlls.

PR #231, opened 2023-09-05
Make PoST Metadata updates atomic

fasmat:


initialization/metadata.go
```diff
@@ -17,14 +19,27 @@ func SaveMetadata(dir string, v *shared.PostMetadata) error {
...
+		return fmt.Errorf("failed to encode metadata: %w", err)
+	}
+
+	if err := tmp.Close(); err != nil {
+		return fmt.Errorf("failed to close tmp file %s: %w", tmp.Name(), err)
+	}
+
+	if err := atomic.ReplaceFile(tmp.Name(), filename); err != nil {
+		return fmt.Errorf("atomic replace file %s with %s: %w", filename, tmp.Name(), err)
 	}
```
poszu: How about using an in-memory `bytes.Buffer` instead of a temporary file?
```go
	buf := new(bytes.Buffer)
	enc := json.NewEncoder(buf)
	enc.SetIndent("", "\t")

	if err := enc.Encode(v); err != nil {
		return fmt.Errorf("failed to encode metadata: %w", err)
	}
	if err := atomic.WriteFile(filepath.Join(dir, MetadataFileName), buf); err != nil { ... }
```

fasmat: Wanted to be efficient, but we can do in memory as well. The file is small and not updated that often.

fasmat: btw. the only difference between `atomic.WriteFile` and `atomic.ReplaceFile` is that `atomic.WriteFile` first writes to a temporary file created by `os.CreateTemp` and then calls `atomic.ReplaceFile(tmp, dest)`.

poszu: Yeah but the code is simpler which is the most important in this case :)

poszu (approved):

PR #241, opened 2023-09-19
Require -numUnits in postcli

### LINKED ISSUE #236: postcli should never delete existing postdata files

lrettig:
Replaces #239 
Closes #236 

This is a much simpler, stopgap fix for this issue that doesn't require changing the lower level post code.

### DIFF
--- a/cmd/postcli/main.go
+++ b/cmd/postcli/main.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strconv"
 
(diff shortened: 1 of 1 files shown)


fasmat (approved):

fasmat:
LGTM

PR #245, opened 2023-10-17
Update post-rs

fasmat:
This updates the post library to use the most recent version of https://github.com/spacemeshos/post-rs

Merge and release after https://github.com/spacemeshos/post-rs/pull/133/


initialization/layout_test.go
```diff
@@ -7,7 +7,6 @@ import (
 )
 
 func TestMaxFileSize(t *testing.T) {
-	t.Parallel()
```
poszu: These tests should be fine to execute in parallel.

poszu (approved):

PR #249, opened 2023-11-23
Bump post-rs to v0.6.1

poszu:
There are a few breaking changes in the C FFI of post-rs library:
- post config is now split into `ProofConfig` and `InitConfig`,
- scrypt is now configured with N, R, and P instead of the factors for a better UX.


internal/postrs/proof.go
```diff
@@ -20,6 +19,26 @@ import (
...
+	K1 uint32 // K1 specifies the difficulty for a label to be a candidate for a proof.
+	K2 uint32 // K2 is the number of labels below the required difficulty required for a proof.
+	K3 uint32 // K3 is the size of the subset of proof indices that is validated.
+
+	PowDifficulty [32]byte
+}
+
+func NewScryptParams(n, r, p uint) ScryptParams {
+	return ScryptParams{
+		n: C.ulong(n),
```
fasmat: What's the reason for long instead of `uint32_t`?

poszu: Actually, it should be `C.size_t`. The type on Rust side is `usize`, which is "pointer type size". Fixed.

fasmat:
We probably need to remove the custom `SDKROOT` on macOS M1 builds to get it working on mac. Windows and Linux seem to be good with this change and https://github.com/spacemeshos/post-rs/pull/155

fasmat (approved):

PR #256, opened 2023-12-22
Distributed verification

poszu:
Changes to support distributed verification (https://github.com/spacemeshos/go-spacemesh/issues/5185):
- added verification modes (_All_, _One_, and _Subset_)
- _Subset_ takes the number of indices to verify and a seed
  *  the seed is used to initialize local randomness for selecting K3 indices (so that nodes verify different subsets)
- return the invalid index in POST proof - required for malfeasance proof

:bulb: There are some changes in C definitions. It's caused by improved enum definitions in the FFI on the post-rs side.


internal/postrs/pos_verifier.go
```diff
@@ -70,14 +71,24 @@ func VerifyPos(dataDir string, scryptParams ScryptParams, o ...VerifyPosOptionsF
...
-	case C.Ok:
+	switch result.tag {
+	case C.VerifyResult_Ok:
 		return nil
-	case C.Invalid:
-		return ErrInvalidPos
-	case C.InvalidArgument:
+	case C.VerifyResult_Invalid:
+		result := castBytes[C.VerifyPosResult_Invalid_Body](result.anon0[:])
+		return fmt.Errorf("%w: file %d contains invalid label at offset %d", ErrInvalidPos, result.file, result.offset)
```
fasmat, line 79: That seems a) dangerous and b) very platform / compiler specific; i.e. just because it works on windows/amd64 doesn't mean it works on macOS/arm64. I think it would be simpler and safer to just return "file n contains invalid label at offset x" as `string` from `post-rs` and then wrap it in the returned error:

```suggestion
	case C.VerifyResult_Invalid:
		return fmt.Errorf("%w: %s", ErrInvalidPos, C.GoString(result.error))
```

internal/postrs/proof.go
```diff
@@ -152,7 +144,49 @@ func (v *Verifier) Close() error {
...
+
+// Verify only the selected index.
+// The `ord` is the ordinal number of the index in the proof to verify.
+func VerifyOne(ord int) VerifyOptionFunc {
+	return func(o *verifyOptions) {
+		// `o.mode` is a C tagged union
+		o.mode.tag = C.VerifyMode_One
+		o.mode.anon0 = (*(*[32]byte)(unsafe.Pointer(&C.VerifyMode_One_Body{
+			index: C.size_t(ord),
+		})))
```
fasmat: I don't think we should serialize C objects unsafely in Go and pass them to Rust via cgo. This seems dangerous and  I think we can do without 🙂

internal/postrs/proof.go
```diff
@@ -152,7 +144,49 @@ func (v *Verifier) Close() error {
...
+func VerifySubset(k3 uint, seed []byte) VerifyOptionFunc {
+	return func(o *verifyOptions) {
+		// `o.mode` is a C tagged union
+		o.mode.tag = C.VerifyMode_Subset
+		o.mode.anon0 = (*(*[32]byte)(unsafe.Pointer(&C.VerifyMode_Subset_Body{
+			k3: C.size_t(k3),
+			seed: C.ArrayU8{
+				ptr: (*C.uchar)(unsafe.SliceData(seed)),
+				len: C.size_t(len(seed)),
+				cap: C.size_t(cap(seed)),
```
fasmat: `cap` in my opinion should never be needed to be passed to rust; most of the time it will be the same as `len` and I hope rust is not modifying `seed`, since we are only passing an unsafe pointer to a variable that is managed by the Go GC?

Why not pass the `seed` with `C.CBytes([]byte)`? Yes it creates a copy but even calling this function 1 Mio. times only copies 32 MB of data on the heap and its safer in a sense that the created byte array is not managed by the GC, can be modified on the Rust side and easily be freed after use with `C.free`?

internal/postrs/proof.go
```diff
@@ -152,7 +144,49 @@ func (v *Verifier) Close() error {
 	return nil
 }
 
-func (v *Verifier) VerifyProof(proof *shared.Proof, metadata *shared.ProofMetadata, logger *zap.Logger, cfg Config, scryptParams ScryptParams) error {
+type verifyOptions struct {
+	mode C.VerifyMode
+}
```
fasmat: I don't think an options struct should contain C types. I think would be safer to use go types in go and only convert them right before calling the actual C function.

proving/proving_options.go
```diff
@@ -67,6 +67,13 @@ func WithPowFlags(flags config.PowFlags) OptionFunc {
 	}
 }
 
+func LightMode() OptionFunc {
+	return func(o *option) error {
+		o.powFlags = config.RecommendedPowFlags()
+		return nil
+	}
+}
```
fasmat, line 75: Why don't we simplify here the same way as for the post service? `fast` and `light` are the only two options. We could just pass a boolean to `post-rs` that if `true` uses `fast` and otherwise `light`. Wdyt?

proving/proving_test.go
```diff
@@ -163,84 +161,10 @@ func Test_Generate_DetectInvalidParameters(t *testing.T) {
...
-			WithPowFlags(postrs.GetRecommendedPowFlags()),
+			LightMode(),
 		)
 		var errConfigMismatch initialization.ConfigMismatchError
 		require.ErrorAs(t, err, &errConfigMismatch)
 		require.Equal(t, "LabelsPerUnit", errConfigMismatch.Param)
 	})
 }
-
-func Test_Generate_TestNetSettings(t *testing.T) {
```
fasmat, line 174: Why did you delete this test?

fasmat (commented):
I'm not a fan of the proposed ABI for `post-rs`. I think it would be much simpler to just have 3 verify functions on the ABI level:

- `C.verifyAll(...)`
- `C.verifyOne(index, ...)`
- `C.verifySubset(k3, seed, ...)`

Outside of the `internal` package we would still have only one `Verify` method with functional options that just depending on the provided options calls a different C function. On the rust side I'm sure that the 3 exported functions can just call one general `verify` function with different parameters depending on the ABI function that was called. This makes the C interface much simpler, no need to serialize objects / structs across ABI if we can do it with simple types.

With arrays I see it similarly:

```go
cSeed := C.CBytes(seed)
defer C.free(cSeed)

// len(seed) is arguably not needed because we can already check on the Go side that it is 32 bytes
// and return an error if it is not
C.verify(..., cSeed, len(seed), ...)
```

seems easier and is safer than:

```go
cSeed := C.ArrayU8{
	ptr: (*C.uchar)(unsafe.SliceData(seed)),
	len: C.size_t(len(seed)),
	cap: C.size_t(cap(seed)),
}
C.verify(..., cSeed, ...)
```

I'm not even sure that the 2nd one is necessarily faster than the first one 🤔

proving/proving_test.go
poszu: Because it's not useful.

poszu:
> I'm not a fan of the proposed ABI for `post-rs`. I think it would be much simpler to just have 3 verify functions on the ABI level:
>    * `C.verifyAll(...)`
>    * `C.verifyOne(index, ...)`
>    * `C.verifySubset(k3, seed, ...)`
> 
> Outside of the `internal` package we would still have only one `Verify` method with functional options that just depending on the provided options calls a different C function. On the rust side I'm sure that the 3 exported functions can just call one general `verify` function with different parameters depending on the ABI function that was called. This makes the C interface much simpler, no need to serialize objects / structs across ABI if we can do it with simple types.

Sure, it's possible to create 3 functions instead. I'm not sure if that's simpler, you trade 1 complexity for another.

> With arrays I see it similarly:
> 
> ```go
> cSeed := C.CBytes(seed)
> defer C.free(cSeed)
> 
> // len(seed) is arguably not needed because we can already check on the Go side that it is 32 bytes
> // and return an error if it is not
> C.verify(..., cSeed, len(seed), ...)
> ```
> 
> seems easier and is safer than:
> 
> ```go
> cSeed := C.ArrayU8{
> 	ptr: (*C.uchar)(unsafe.SliceData(seed)),
> 	len: C.size_t(len(seed)),
> 	cap: C.size_t(cap(seed)),
> }
> C.verify(..., cSeed, ...)
> ```
> 
> I'm not even sure that the 2nd one is necessarily faster than the first one 🤔

Avoiding a copy would be faster, but `ArrayU8` is not about performance. Originally, it was required to be able to pass an object allocated on the Rust side to Go and to be able to free it afterward (hence it must contain `cap`). It's not required to pass data from Go to Rust - it was used as a convenience as the type was already there.

Additionally, in situations when it is needed to pass 2 slices, it's better to pass 2 objects that contain ptr + len rather than ptr0, len0, ptr1, len1 as the former is less error-prone.

internal/postrs/pos_verifier.go
poszu: All C code is dangerous :upside_down_face:. What you propose works for verifying POS, but wouldn't work for verifying proof - we need the index there.

proving/proving_options.go
poszu: _Fast_ and _Light_ are **not** the only two options. There could be other combinations, for example setting LARGE_PAGES, which we don't do but could _slightly_ improve performance on some setups (but won't work on most).

I wanted to keep the post-rs API as flexible as possible so it exposes flags as is. Sure, we could change it but not in this PR :pray:.

internal/postrs/proof.go
poszu: That's how C tagged enum works. The structure has a tag and a container to fit the largest variant (CGO decided to call it `anon0` (anonymous union)). The container literally contains the memory of the variant as in C. The layout is defined (https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md) and not "platform specific". IMHO, this isn't any more unsafe from other unsafe operations we do. It may look scary though :).

internal/postrs/proof.go
poszu: Changed to 3x `verify_x()` functions and passing a copy of the seed as argument.

internal/postrs/proof.go
poszu: Updated

internal/postrs/proof.go
poszu: Anyway - I split `verify_proof()` into three functions so the C tagged union is not needed on Go -> C direction anymore. However, I kept using `VerifyResult` enum in result of `verify_x()` functions.

internal/postrs/pos_verifier.go
fasmat: I think the issue I have with this approach is that the return value of the verify function is essentially a `void *` (or at least of the field `anon0` in the returned object), but I don't think it needs to be. Neither here nor in the case of verifying proof. The return value for `C.verify_pos` can just be (in go syntax):

```go
type pos_result struct {
	tag    int    // the error code
	file   string // only set when tag == C.VerifyResult_Invalid
	offset int    // only set when tag == C.VerifyResult_Invalid
}
```

and the return value for `C.verify_proof` can instead be:

```go
type proof_result struct {
	tag      int // the error code
	index_id int // only set when tag == C.VerifyResult_InvalidIndex
}
```

Without a generic approach the object-byte-cast wouldn't be necessary 🙂.

fasmat: Maybe we can just give the `anon0` field a name? Then I'd already be happy 🙂

fasmat (approved):

internal/postrs/pos_verifier.go
poszu: The header is autogenerated with https://github.com/mozilla/cbindgen. I looked but I didn't find a way to force it to name the union :(

PR #261, opened 2024-01-09
Run build and tests on GH runner instead of self hosted Mac M1

fasmat:


poszu (approved):

fasmat:
To reduce costs we stick with self hosted for now

PR #262, opened 2024-01-11
Add docker image for postcli

fasmat:
This adds a docker image for `postcli`. Since within docker it probably cannot access the GPU its intended use is primarily for system tests.

Also extended `postcli` to allow to change Scrypt `N`. Changing any settings that break compatibility with mainnet (or would delete data) require explicit approval with `-yes`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,54 @@
+FROM golang:1.21 as builder
+RUN set -ex \
+    && apt-get update --fix-missing \
+    && apt-get install -qy --no-install-recommends \
+    unzip sudo \
+    ocl-icd-opencl-dev
+
+WORKDIR /src
+COPY Makefile* .
+RUN make get-postrs-lib
+
+# We want to populate the module cache based on the go.{mod,sum} files.
+COPY go.mod .
+COPY go.sum .
+
+RUN go mod download
+
+# Here we copy the rest of the source code
+COPY . .
+
+# And compile the project
+RUN --mount=type=cache,id=build,target=/root/.cache/go-build make build
+
+FROM ubuntu:22.04 AS postcli
+ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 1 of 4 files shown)


cmd/postcli/main.go
```diff
@@ -138,6 +129,24 @@ func processFlags() error {
 		return fmt.Errorf("invalid commitmentAtxId: %w", err)
 	}
 
+	if flagSet["labelsPerUnit"] && (cfg.LabelsPerUnit != config.MainnetConfig().LabelsPerUnit) {
+		fmt.Println("WARNING: labelsPerUnit is set to a non-default value. This makes the initialization incompatible " +
+			"with mainnet. If you're trying to initialize for mainnet, please remove the -labelsPerUnit flag")
+		if !yes {
+			fmt.Println("If you're sure you want to continue, please run with -yes")
+			os.Exit(1)
+		}
```
poszu: Maybe it could prompt for `y/N` if `-yes` is not set, similarly to other tools like apt etc.?

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+    ocl-icd-libopencl1 clinfo \
+    pocl-opencl-icd libpocl2 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
+    && locale-gen en_US.UTF-8 \
+    && update-locale LANG=en_US.UTF-8 \
+    && echo "$TZ" > /etc/timezone
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US.UTF-8
+ENV LC_ALL en_US.UTF-8
```
poszu: Is it all needed? I think only the OpenCL runtime is needed.

fasmat: I removed most dependencies

poszu (approved):

PR #263, opened 2024-01-22
Add option to keep redundant files

xearl4:
A long-standing pain point for go-spacemesh users is that it simply deletes postdata files that are deemed "redundant". For a go-spacemesh managed through smapp, the behaviour may be reasonable and comprehensible. For CLI users, however, this often means that a simple config mistake of setting the wrong num-units in the config file leads to weeks of postdata file initialisation being quickly wiped out.

With this change, the current default behaviour is preserved: redundant files are deleted. If the new option is enabled, an info entry is logged for redundant files, but they are not deleted.

This change prepares the stage for a dependent change in go-spacemesh.

### DIFF
--- a/initialization/initialization.go
+++ b/initialization/initialization.go
@@ -61,6 +61,8 @@ type option struct {
 	logger            *Logger
 	powDifficultyFunc func(uint64) []byte
 	referenceOracle   *oracle.WorkOracle
+
+	keepRedundantFiles bool
 }
 
 func (o *option) validate() error {
@@ -135,6 +137,13 @@ func WithLogger(logger *zap.Logger) OptionFunc {
 	}
 }
 
+func WithKeepRedundantFiles() OptionFunc {
+	return func(opts *option) error {
+		opts.keepRedundantFiles = true
+		return nil
+	}
+}
+
 // withDifficultyFunc sets the difficulty function for the initializer.
 // NOTE: This is an internal option for tests and should not be used by external packages.
 func withDifficultyFunc(powDifficultyFunc func(uint64) []byte) OptionFunc {
@@ -182,6 +191,8 @@ type Initializer struct {
 	logger            *Logger
 	referenceOracle   *oracle.WorkOracle
 	powDifficultyFunc func(uint64) []byte
+
+	keepRedundantFiles bool
 }
 
 func NewInitializer(opts ...OptionFunc) (*Initializer, error) {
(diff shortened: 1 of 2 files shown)


schinzelh:
Fixes https://github.com/spacemeshos/go-spacemesh/issues/4594

initialization/initialization.go
```diff
@@ -295,7 +307,7 @@ func (init *Initializer) Initialize(ctx context.Context) error {
 		zap.Int("firstFileIndex", layout.FirstFileIdx),
 		zap.Int("lastFileIndex", layout.LastFileIdx),
 	)
-	if err := removeRedundantFiles(init.cfg, init.opts, init.logger); err != nil {
+	if err := checkRedundantFiles(init.cfg, init.opts, init.keepRedundantFiles, init.logger); err != nil {
 		return err
 	}
```
fasmat, line 312: Instead of passing the `bool` to the function I'd prefer to instead not call `removeRedundantFiles` when it is `true`:

```go
if !init.keepRedundantFiles {
	if err := removeRedundantFiles(init.cfg, init.opts, init.logger); err != nil {
		return err
	}
}
```

no need to iterate over all files telling the user what is all ignored when the user actively chose to ignore files.

fasmat (commented):
Thanks for your contributions! 🙂

`postcli` in `post/cmd/postcli` might also benefit from the new functional option of the Initializer, if you find some time to take a look at it.

You might want to consider adding the `KeepRedundantFiles` flag to `InitOpts` in `post/config/config.go` instead of using a dedicated functional option for it. This will probably make integration in `go-spacemesh` easier.

Setting the flag via the cmd line and/or config is a bit involved - I suggest you add the bool to `PostSetupOpts` in `go-spacemesh/activation/post.go` which is already passed to the service that uses the `Initializer` to generate the post data - `PostSetupManager` in `go-spacemesh/activation/post.go`. `PostSetupOpts` also has a method to convert it to `config.InitOpts` used by the `Initializer` which you can extend with the new flag.

If you want to expose the flag not only via the node-config.json but also via cmdline parameters you can take a look at `go-spacemesh/cmd/root.go` to see how other cmd line parameters are handled. 🙂

xearl4:
> `postcli` in `post/cmd/postcli` might also benefit from the new functional option of the Initializer, if you find some time to take a look at it.

Sure. Main question is: what do we want the default to be? Should `postcli` default to deleting, and keep files if you pass a CLI flag -- or should it keep redundant files by default, and only delete when passed, say, `--cleanup`?

initialization/initialization.go
xearl4: Certainly fine as well.

But in the usage scenarios I envision, many users will set "keep" as a default everywhere, and then still find it helpful to see log messages where additional files are discovered, to assist them in keeping track of things. The log message helps in cases where one forgot to adjust the num-units after an increase; it also helps in cleaning up after a decrease: it reminds the user to think about a situation that was previously handled in a one-sided manner by the program by deleting files.

In fact, I originally had exactly the variant you proposed, with no log message at all :) I only added the log message because I thought that might make it easier to upstream. Having used it while resizing (mostly increasing) lots of postdata over the past few weeks, I have come to slightly prefer the log message.

All that said: either way is fine with me.

fasmat:
`postcli` requires explicitly stating `numUnits`; a user specifying fewer units than they have on disk thereby probably intends to initialize the value they provided.

On the other hand having a `-keepFiles` flag makes less sense then having a `-deleteFiles` flag. So I guess the default should probably be to not delete 🤔

pigmej:
I agree with @fasmat  here.

by default, it should not delete UNLESS specified.

postcli case is a bit specific because we support `from file` etc., so trying to guess what the user wanted is, IMO tricky; we should basically do what the user said us to do, so whatever is specified -> ok.

fasmat:
Hi @xearl4 @schinzelh,

we just merged a change that adds additional checks to `postcli` to prevent users from accidentally corrupting/deleting their files: https://github.com/spacemeshos/post/pull/270 and released the changed version as `v0.12.0` (pre-release for now).

With the upcoming `v1.4.x` version of `go-spacemesh` the post data directory can be set to read only after initialization to prevent accidental deletion / corruption of existing data and with the upcoming multi-smeshing release we plan to remove initialization from the node completely and only use `postcli` for creating PoST data.

I will therefore close this PR. If you feel additional changes are needed feel free to re-open and update your PR and re-start the discussion 🙂

xearl4:
Sounds good, thanks!

PR #270, opened 2024-02-26
Update postcli for compatibility with multismeshing

### LINKED ISSUE #219: postcli should use postdata_metadata.json for default values
Postcli now contains many functions beyond simply generating post data.

When -datadir is provided and postdata_metadata.json exists the metadata should take priority over postcli defaults. Requiring the user to supply values each time complicates things and introduces scope for error.

Where the user does provide an explicit argument to postcli which clashes with a postdata_metadata.json value a warning should be thrown as is the case with an unexpected -id argument.

- commitmentAtxId
- id
- labelsPerUnit
- maxFileSize
- numUnits

fasmat:
Part of https://github.com/spacemeshos/go-spacemesh/issues/5089

Closes #219 

This updates `postcli` with the changes for multi-smeshing setup.

- `postcli` now stores a generated key in `identity.key` instead of `key.bin` and informs the user that they have to copy the file to the `datadir` of the node
- Added multiple checks to ensure users don't accidentally corrupt or delete their PoST data files:
  - the key in `identities.key` is checked to be compatible with the `NodeID` in the `postdata_metadata.json` file (only if both files are present
  - a provided `-id` is checked against both the `identities.key` file and `NodeID` in `postdata_metadata.json` and user has to confirm a warning that providing a mismatched value potentially corrupts their data.
  - `-numUnits` has to be set if `postdata_metadata.json` cannot be found. If `postdata_metadata.json` is present the value will be read from there. If the provided value does not match the value in the `postdata_metadata.json` file the user has to confirm that they want to continue and is informed that providing a lower value will delete data.
  - `-commitmentAtxId` has to be set if `postdata_metadata.json` cannot be found. If `postdata_metadata.json` is present the value will be read from there. If the provided value does not match the value in the `postdata_metadata.json` file the user has to confirm that they want to continue and is informed that providing a wrong value can corrupt their data.
  - `provider` has to be set
  - `labelsPerUnit` and `scryptN` trigger a warning if set and do not match mainnet values. User has to confirm that this will generate data that is not compatible with mainnet.
- All text that is printed now uses `log.Print` and `log.Fatal` instead of `fmt.Println`


cmd/postcli/main.go
```diff
@@ -102,75 +100,144 @@ func askForConfirmation() {
...
+		} else if idHex != hex.EncodeToString(key) {
+			log.Println("WARNING: it appears that", opts.DataDir, "was previously initialized with a generated key.")
+			log.Println("The provided id does not match the generated key.")
+			log.Println("\tCurrent value:", hex.EncodeToString(key))
+			log.Println("\tValue passed to postcli:", idHex)
+			if !yes {
+				log.Println("CONTINUING MIGHT ERASE EXISTING DATA. MAKE ABSOLUTELY SURE YOU SPECIFY THE CORRECT VALUE.")
+			}
+			askForConfirmation()
+		}
```
poszu: I think it would be best to just fail in this situation rather than to allow users to continue. It has no chance of working.

cmd/postcli/main.go
```diff
@@ -102,75 +100,144 @@ func askForConfirmation() {
...
+		if meta != nil && !bytes.Equal(commitmentAtxId, meta.CommitmentAtxId) {
+			log.Println("WARNING: it appears that", opts.DataDir, "was previously initialized with a different `commitmentAtxId` value.")
+			log.Println("\tCurrent value:", hex.EncodeToString(meta.CommitmentAtxId))
+			log.Println("\tValue passed to postcli:", commitmentAtxIdHex)
+			if !yes {
+				log.Println("CONTINUING MIGHT ERASE EXISTING DATA. MAKE ABSOLUTELY SURE YOU SPECIFY THE CORRECT VALUE.")
+			}
+			askForConfirmation()
+			commitmentAtxIdHex = hex.EncodeToString(meta.CommitmentAtxId)
+		}
```
poszu: Same here - I think it's better to fail.

cmd/postcli/main.go
```diff
@@ -344,19 +410,27 @@ func cmdVerifyPos(opts config.InitOpts, fraction float64, logger *zap.Logger) {
...
-	pub := ed25519.NewKeyFromSeed(dst[:ed25519.SeedSize]).Public().(ed25519.PublicKey)
+	return ed25519.NewKeyFromSeed(dst[:ed25519.SeedSize]).Public().(ed25519.PublicKey), nil
+}
+
+func cmdVerifyPos(opts config.InitOpts, fraction float64, logger *zap.Logger) {
+	log.Println("cli: verifying", edKeyFileName)
+	pub, err := loadKey()
+	if err != nil {
+		log.Fatalf("failed to load public key from %s: %s\n", edKeyFileName, err)
+	}
```
poszu: Now that I think of this, it could just read the key from the metadata JSON file. There is no reason to require the identity.key to exist in order to verify the POS data.

fasmat: I changed the code to skip verification of the key when it is missing; but left the check if a key is present to give a user the option to still verify it against the initialised data (if i.e. they forgot which key belongs to which PoST).

cmd/postcli/main.go
```diff
@@ -416,23 +406,25 @@ func loadKey() (ed25519.PublicKey, error) {
 func cmdVerifyPos(opts config.InitOpts, fraction float64, logger *zap.Logger) {
 	log.Println("cli: verifying", edKeyFileName)
 	pub, err := loadKey()
-	if err != nil {
+	switch {
+	case errors.Is(err, os.ErrNotExist):
+		log.Println("cli:", edKeyFileName, "does not exist. Skipping verification")
```
poszu: The message is a little misleading. It skips cross-checking the keys in the identity key file and metadata JSON file, not verifying the data. Perhaps just don't print anything?
```suggestion
```

poszu (approved):

PR #275, opened 2024-03-06
Store key as `<pubkey>.key`

fasmat:
Related to https://github.com/spacemeshos/go-spacemesh/pull/5643

Instead of `<identity>.key` store the key with the public key as the file name (abbreviated).

### DIFF
--- a/cmd/postcli/main.go
+++ b/cmd/postcli/main.go
@@ -26,8 +26,6 @@ import (
 	"github.com/spacemeshos/post/verifying"
 )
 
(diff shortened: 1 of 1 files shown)


poszu (approved):
Could you also update the readme: https://github.com/spacemeshos/post/blob/fdd5647948fe150cb88a279f6e54e673d9e3aa0d/cmd/postcli/README.md#L101-L107

fasmat:
Closing as having "random" key names might be confusing and the change makes the code more convoluted than necessary.

Issue #274, opened 2024-03-03
latest develop version doesn't work on rtx 3080


kiss81:
I built the latest develop postcli version. I can confirm it solves the 100% cpu issue on nvidia. It only works on my 4070ti somehow. My 3080 (non ti) crashes...
xubuntu 23.10
nvidia driver 550.54.14

`2024-03-03T14:34:38.640+0100    INFO    selecting 1 provider from 2 available    {"module": "scrypt_ocl", "file": "scrypt-ocl/src/lib.rs", "line": 354}
2024-03-03T14:34:38.640+0100    INFO    Using provider: [GPU] NVIDIA CUDA/NVIDIA GeForce RTX 3080    {"module": "scrypt_ocl", "file": "scrypt-ocl/src/lib.rs", "line": 367}
2024-03-03T14:34:38.640+0100    INFO    device memory: 9997 MB, max_mem_alloc_size: 2499 MB, max_compute_units: 68, max_wg_size: 1024    {"module": "scrypt_ocl", "file": "scrypt-ocl/src/lib.rs", "line": 153}
thread '<unnamed>' panicked at /home/gh-action-runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ocl-0.19.6/src/standard/pro_que.rs:480:17:
assertion failed: ctx.devices().contains(&device)
note: run with RUST_BACKTRACE=1 environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
SIGABRT: abort
PC=0x775827699a1b m=12 sigcode=18446744073709551610
signal arrived during cgo execution`

`goroutine 1 gp=0xc0000061c0 m=12 mp=0xc000101008 [syscall]:
runtime.cgocall(0x5de978, 0xc00002c860)
    /usr/local/go/src/runtime/cgocall.go:157 +0x4b fp=0xc00002c838 sp=0xc00002c800 pc=0x40712b`

kiss81:
Update from multiple users on discord:
It doesn't matter what gpu. It's just the fact there is more then one gpu installed. The first works, but second one will give an error.

kiss81:
Tried the latest postcli 12.1 on my second gpu. No luck either

_ocl", "file": "scrypt-ocl/src/lib.rs", "line": 153}
thread '<unnamed>' panicked at /home/gh-action-runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ocl-0.19.6/src/standard/pro_que.rs:480:17:
[...]

PR #277, opened 2024-03-12
Fix scrypt close

poszu:
Fix segfaults when closing a `postrs.Scrypt` instance that first failed creation. Additionally, fix a deadlock in UT on platforms with more than 1 GPU.


fasmat (approved):

fasmat:
Does this fix https://github.com/spacemeshos/post/issues/274 ?

poszu:
> Does this fix #274 ?

No, this fixes the crash on the Go side during `Close()` that happened when creating the initializer failed.

PR #282, opened 2024-03-20
Fix passing invalid log level to postrs logger

poszu:
The zap noop logger has `zapcore.InvalidLevel` log level, which is not represented in the go -> C log levels mapping. The current code, for the noop logger, passes a default Level value of 0. This is not valid and causes undefined behavior on the Rust side, which started to cause a crash since Rust 1.76.

The proper behavior is to not set the logger in such case.

### DIFF
--- a/internal/postrs/log.go
+++ b/internal/postrs/log.go
@@ -42,13 +42,19 @@ var (
 )
 
 func setLogCallback(logger *zap.Logger) {
-	oncer.Do(func() {
-		C.set_logging_callback(levelMap[logger.Level()], C.callback(C.logCallback))
-	})
+	level, ok := levelMap[logger.Level()]
+	if !ok {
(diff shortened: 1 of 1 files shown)


fasmat:
I think we should reconsider not using a global on the Rust side for logging. The logger on the Go side is not a global which especially in tests causes issues. `setLogCallback` is called multiple times but the oncer inside only once meaning all tests for instance will have the same logging level (as set by the first) in a testrun.

poszu:
> I think we should reconsider not using a global on the Rust side for logging. The logger on the Go side is not a global which especially in tests causes issues. `setLogCallback` is called multiple times but the oncer inside only once meaning all tests for instance will have the same logging level (as set by the first) in a testrun.

Maybe, but that's not in the scope of this fix.

fasmat (approved):

PR #293, opened 2024-06-13
Encode nodeid in hex for invalid vrf nonce error

poszu:
Currently the error text is like:
```
nonce 0 is not valid for node [53 165 217 54 93 19 23 195 0 40 56 18 100 9 225 56 50 28 87 165 101 29 117 132 133 51 108 30 126 90 241 1]
```

After this change its like:
```
nonce 0 is not valid for node 35a5d9365d1317c3002838126409e138321c57a5651d758485336c1e7e5af101
```

### DIFF
--- a/verifying/verifying.go
+++ b/verifying/verifying.go
@@ -52,7 +52,7 @@ func VerifyVRFNonce(nonce *uint64, m *shared.VRFNonceMetadata, opts ...OptionFun
 	}
 
 	if res.Nonce == nil || *res.Nonce != *nonce {
-		return fmt.Errorf("nonce %v is not valid for node %v", *nonce, m.NodeId)
+		return fmt.Errorf("nonce %v is not valid for node %x", *nonce, m.NodeId)
 	}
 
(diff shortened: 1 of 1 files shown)


verifying/verifying.go
```diff
@@ -52,7 +52,7 @@ func VerifyVRFNonce(nonce *uint64, m *shared.VRFNonceMetadata, opts ...OptionFun
 	}
 
 	if res.Nonce == nil || *res.Nonce != *nonce {
-		return fmt.Errorf("nonce %v is not valid for node %v", *nonce, m.NodeId)
+		return fmt.Errorf("nonce %v is not valid for node %x", *nonce, m.NodeId)
```
fasmat, line 55: `types.NodeID` has a `String()` method that prints the nodeID in short form.

```suggestion
		return fmt.Errorf("nonce %v is not valid for node %s", *nonce, m.NodeId)
```

fasmat (approved):

PR #297, opened 2024-07-03
Remove duplicated LDFLAGS definitions

fasmat:
This should get rid of the warnings during macOS builds and tests.


poszu (dismissed):

fasmat:
Abandoning for now. The warning are annoying but getting rid of them seems not to be worth the effort.

